PR #44, opened 2023-05-19
Catching "Already Incepted" error

rodolfomiranda:
While booting when the Client AID is already incepted, KERIA respond with "500 Internal Server Error"
This PR changes that behavior by catching the error and responding with error code 400 and the description received from the upper class that in this case is "Already incepted".

I assigned error 400 because I couldn't find an error more descriptive. There's an error code 409 "Conflicting" that can be use if it's considered more appropriate.

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -543,7 +543,11 @@ class BootEnd:
 
         caid = icp.pre
 
-        agent = self.agency.create(caid=caid)
+        try:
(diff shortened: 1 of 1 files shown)


pfeairheller (changes requested):
A better approach here would be to check for the existence of icp.pre in the agent.hby.kevers and fail if its there.  

Also, please add test coverage.

pfeairheller:
I'm going to push another PR that has the suggested fix and tests in it.

PR #58, opened 2023-06-20
Add registry long running Op handler

kentbull:
Locates the registry operation by anchor and returns its state whether done or not done.


pfeairheller:
This code has been moved to a broader PR

PR #67, opened 2023-07-07
Interceptor

AlexAndrei98:
I wanted to create a new class that allows to push events that are happening inside the cloud agent to other backend processes. 
It is similar to the notifier class but it is used to "notify" other web services.
I would like to notify the web services of the following event:
simple aid inception
simple aid rotation
simple aid delegation
multisig aid inception
multisig aid rotation
multisig aid delegation
multisig request 
multisig signatures
acdc issuance
acdc revocation
acdc presentation

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -245,6 +245,8 @@ class Agent(doing.DoDoer):
         self.caid = caid
         if interceptor_webhook is not None:
             self.interceptor = Interceptor(interceptor_webhook, interceptor_headers)
+        else:
+            self.interceptor = None
 
         self.swain = delegating.Boatswain(hby=hby)
         self.counselor = Counselor(hby=hby)
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -123,6 +123,7 @@ class Agency(doing.DoDoer):
         self.cf = None
         self.intercepts = decking.Deck()
         doers = []
+        self.interceptor = None
         if interceptor_webhook is not None:
             self.interceptor = InterceptorDoer(interceptor_webhook, interceptor_headers, cues=self.intercepts)
             doers.append(self.interceptor)
@@ -415,7 +416,7 @@ class Messager(doing.Doer):
 
 class Witnesser(doing.Doer):
 
-    def __init__(self, receiptor, witners, intercepts):
+    def __init__(self, receiptor, witners, intercepts=None):
         self.receiptor = receiptor
         self.witners = witners
         self.intercepts = intercepts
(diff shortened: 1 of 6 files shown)


pfeairheller:
I agree this is good functionality to add.  When we built kiwiing and the Keep we struggled with the need for "notifications" of agent activity (transient events that are only important if a client was currently running) and user notifications that needed to persist until the user acknowledged and dealt with them (for example, a credential was issued, a presentation was requested, a multisig group was proposed).  I'd like us to make sure we don't conflate these two use cases in KERIA.

This functionality seems like a separate type of notification all together.  Notifying some process controlled by the host of the KERIA service of all activity.  Seems like KERIA doesn't need to worry about persistence of these notifications and only fire and forget if and only if someone is currently listening (no retries, etc).

Instead of configuring headers to be passed how about we just sign every message from the Agent that generates the message.  That should be all the information a receiving process would need to authenticate the request and handle it.  

Also, this feels a lot like instrumentation and metric processing for a KERIA service so should we consider using something like Prometheus to generate events and allow consumers to use all the tools available from that ecosystem to consume, process and present this data?  That will make signing the data a little more challenging (and maybe not needed) so worth further discussion.

Regardless of path, we'll need to the tests to pass... 😉

AlexAndrei98:
Adding Prometheus for the endpoints I believe is the easiest path to gather performance metrics! However, like you said, the interceptor is interested in pushing low level Keri events to an internal queue. 

I think we can add signify headers but some backends still work with bearer tokens so that's my reasoning on why the headers would be needed.

AlexAndrei98:
@pfeairheller What are some good places to start putting some interceptor? At first I was thinking in the endpoint definition but then I realized that some async operations you would only find out if the client pulls so I started looking through the escrow porcesses like 
[here](https://github.com/WebOfTrust/keria/blob/main/src/keria/app/credentialing.py#L742), or  [here](https://github.com/WebOfTrust/keria/blob/main/src/keria/app/credentialing.py#L742) and [here](https://github.com/WebOfTrust/keria/blob/main/src/keria/app/credentialing.py#L948) . I see that the messages and attachments(like signagures) are bytearrays, can I just go ahead and base64 encode them or parse each of them and create a pretty() version of the object?

src/keria/app/agenting.py
```diff
@@ -308,15 +312,15 @@ def __init__(self, hby, rgy, agentHab, agency, caid, **opts):
                                      vry=self.verifier)
 
         doers.extend([
-            Initer(agentHab=agentHab, caid=caid),
+            Initer(agentHab=agentHab, caid=caid, metrics = self.agency.metrics),
```
pfeairheller, line 315: This is the list of doers, you have to include `self.interceptor` in this list to get the recur method to be called as a background process.

AlexAndrei98: Can I not put it when I Initialize the agency `super(Agency, self).__init__(doers=[self.interceptor], always=True)` on Line 132 instead of having a logger per agent, I was thinking that since we have a queue per agency instead of agent

AlexAndrei98: @pfeairheller I also tried passing `self.agency.interceptor` but it still not running. Does Interceptor need to be Just a Doer Since Agent is a DoDoer?

pfeairheller:
Closing this old PR due to lack of activity.

PR #69, opened 2023-07-09
Fix VC issuance with witnesses

rodolfomiranda:
Fix the issuance of VC when the issuer has witnesses.

- add witDoers in the list of doers
- add the msgs to witDoers
- fix sequence numbers in op monitor

### DIFF
--- a/src/keria/app/aiding.py
+++ b/src/keria/app/aiding.py
@@ -489,16 +489,16 @@ class IdentifierResourceEnd:
         if hab.kever.delegator:
             agent.anchors.append(dict(alias=name, pre=hab.pre, sn=0))
             op = agent.monitor.submit(hab.kever.prefixer.qb64, longrunning.OpTypes.delegation,
-                                      metadata=dict(pre=hab.pre, sn=hab.kever.sn))
+                                      metadata=dict(pre=hab.pre, sn=serder.sn))
             return op
 
-        if hab.kever.wits:
+        elif hab.kever.wits:
             agent.witners.append(dict(serder=serder))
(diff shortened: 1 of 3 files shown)


src/keria/app/aiding.py
```diff
@@ -489,16 +489,16 @@ def rotate(agent, name, body):
         if hab.kever.delegator:
             agent.anchors.append(dict(alias=name, pre=hab.pre, sn=0))
             op = agent.monitor.submit(hab.kever.prefixer.qb64, longrunning.OpTypes.delegation,
-                                      metadata=dict(pre=hab.pre, sn=hab.kever.sn))
+                                      metadata=dict(pre=hab.pre, sn=serder.sn))
             return op
 
-        if hab.kever.wits:
+        elif hab.kever.wits:
```
pfeairheller, line 495: Since we return out of the previous if, the elif is not needed

src/keria/app/aiding.py
```diff
@@ -489,16 +489,16 @@ def rotate(agent, name, body):
...
             agent.witners.append(dict(serder=serder))
             op = agent.monitor.submit(hab.kever.prefixer.qb64, longrunning.OpTypes.witness,
-                                      metadata=dict(sn=hab.kever.sn))
+                                      metadata=dict(sn=serder.sn))
             return op
-
-        op = agent.monitor.submit(hab.kever.prefixer.qb64, longrunning.OpTypes.done,
+        else:
```
pfeairheller, line 500: Same here

rodolfomiranda: elif/else removed

PR #72, opened 2023-07-12
Fix credential presentation method

rodolfomiranda:
Two small changes added:
- load presenting endpoints
- modify method signature format to match the one used in challenge response


rodolfomiranda:
need to fix the tests.

pfeairheller (approved):

PR #75, opened 2023-07-16
Fix resolving OOBIs not in kever

rodolfomiranda:
That PR fix a`KeyError` error when resolving OOBIs that the `CID` was not on the local never.

Also, I remove a duplicated line. But not sure if it was there on purpose. @pfeairheller ?

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -823,7 +823,6 @@ class OOBICollectionEnd:
                 obr.oobialias = body["oobialias"]
 
             agent.hby.db.oobis.pin(keys=(oobi,), val=obr)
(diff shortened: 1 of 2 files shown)


src/keria/app/agenting.py
```diff
@@ -823,7 +823,6 @@ def on_post(req, rep):
                 obr.oobialias = body["oobialias"]
 
             agent.hby.db.oobis.pin(keys=(oobi,), val=obr)
-            agent.hby.db.oobis.pin(keys=(oobi,), val=obr)
```
rodolfomiranda, line 826: this is the duplicated line, that I'm not sure if it's there on purpose

pfeairheller (approved):

Issue #54, opened 2023-06-13
Crash when no witness endpoint


rodolfomiranda:
If an AID is incepted with a witness that does not respond, it crashes with `ERR: unable to query witness BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha, no http endpoint`

rodolfomiranda:
Crashing in this line: https://github.com/WebOfTrust/keria/blob/62a344b6d5d953244fe0ceed51579519be079620/src/keria/app/agenting.py#L386

I don't know what to `yield` if we catch the exception

lenkan:
@rodolfomiranda was this fixed by #92

PR #78, opened 2023-07-19
Check if witnesses are known

rodolfomiranda:
This PR is an attempt to fix the error reported on issue https://github.com/WebOfTrust/keria/issues/54 by checking if the url is in the `hab` or if the obbi was already resolved for new witnesses introduced in a `icp` or a `rot` event. 

@pfeairheller , if you thinks that this is a correct approach, I will add and fix some tests.


rodolfomiranda:
replaced by PR #92

PR #84, opened 2023-08-03
Sefl contained dockerfile for keria

tibor19:


pfeairheller:
There is already a dockerfile in the images directory

pfeairheller:
Closing because we already have a dockerfile.

PR #91, opened 2023-08-11
Support for https endpoints in witnesses, controllers and agents

rodolfomiranda:


pfeairheller:
See my comment to https://github.com/WebOfTrust/keripy/pull/555.  Same applies here

rodolfomiranda:
This PR is in line with the merged PR on `keripy` https://github.com/WebOfTrust/keripy/pull/555

psteniusubi:
Hi @rodolfomiranda 
I just tried this fix with some smallish issues. 
My https witness endpoints have no port, for example `https://witness-5642.io.ubidemo1.com`
This causes keri to generate invalid oobi urls with `:None` as port, like this `https://witness-5642.io.ubidemo1.com:None`

rodolfomiranda:
YEs, you are correct. We faced the same issues. I'll add that change as well.

psteniusubi:
It looks like this issue exists in many places of keripy too. 
I think it would be simpler to generate urls with [urljoin](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin), using the configured url as base.
Thanks!

rodolfomiranda:
I'm replacing this PR with #103

PR #92, opened 2023-08-11
Check if witnesses are known (copy)

rodolfomiranda:
This is a replacement  PR for #78 because I had  some conflicts when rebasing to the new `development` branch. The other PR will be closed.


pfeairheller (changes requested):
You made the initialization insecure by removing the verification of the request.  That must be fixed before considering this PR

src/keria/app/aiding.py
```diff
@@ -155,9 +155,6 @@ def on_put(self, req, rep, caid):
         ctrlHab = agent.hby.habByName(caid, ns="agent")
         ctrlHab.rotate(serder=rot, sigers=[coring.Siger(qb64=sig) for sig in sigs])
 
-        if not self.authn.verify(req):
```
pfeairheller, line 158: This should not have been removed.

rodolfomiranda: The deletion of those two lines had no sense in that PR. I added again. I probably removed while adding and trying  ne tests.

PR #97, opened 2023-08-31
Simple healthcheck

kentbull:
Useful for Kubernetes readinessProbe and livenessProbe


pfeairheller (changes requested):
Fix the broken tests and this will be good to go.  A simple "is alive" on the boot interface is fine for now.

kentbull:
#144 addresses this

PR #102, opened 2023-09-07
Add TLS support

kentbull:
This adds createHttpServer to create an HTTPS server with `hio.core.tcp.ServerTls`.

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -7,7 +7,7 @@ keria.app.agenting module
 import json
 import os
 from dataclasses import asdict
-from urllib.parse import urlparse
+from urllib.parse import urlparse, urljoin
 
 from keri import kering
 from keri.app.notifying import Notifier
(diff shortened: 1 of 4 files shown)


src/keria/app/agenting.py
```diff
@@ -108,6 +110,31 @@ def setup(name, bran, adminPort, bootPort, base='', httpPort=None, configFile=No
     return doers
 
 
+def createHttpServer(port, app, keypath, certpath, cafilepath):
```
m00sey, line 113: This method probably needs a test to get the patch coverage up

m00sey (changes requested):

src/keria/app/agenting.py
kentbull: I'll add one.

src/keria/app/agenting.py
```diff
@@ -93,7 +95,7 @@ def setup(name, bran, adminPort, bootPort, base='', httpPort=None, configFile=No
         ending.loadEnds(agency=agency, app=happ)
         indirecting.loadEnds(agency=agency, app=happ)
 
-        server = http.Server(port=httpPort, app=happ)
+        server = createHttpServer(httpPort, happ, keypath, certpath, cafilepath)
```
pfeairheller, line 98: If you are going to change this to HTTPS, you will need to change the way you expose the OOBIs so that they are also HTTPs.  Otherwise others will not be able to resolve the oobis for this Agent.

This is brings to mind that we might need multiple cert options because users will probably want to expose this endpoint to the world on one domain name and expose the Admin API on another domain name.  Heck, we may need 3.

pfeairheller (changes requested):
We have to push this support all the way down to OOBI generation for the Agent OOBIs.  See my comment below.

src/keria/app/agenting.py
kentbull: Does Rodolfo's PR #91 address this need?

pfeairheller: No it does not.  You will have to update OobiResourceEnd.on_get to change what scheme (currently hardcoded to http) is used when generating OOBIs, based on the configuration of these flags.  Endpoints that are returned when OOBIs are resolved are controlled by configuration files so they are fine, but not OOBIs themselves.

kentbull: I just hit this problem. I'm getting a `socket.gaierror: [Errno -2] Name does not resolve` when resolving an HTTPS OOBI.
I'll get a PR submitted on this ASAP.

kentbull:
To get the patch coverage up higher I need to add test coverage for the `keria start` command. In order to do that it seemsI need to actually start a KERIA server instance which means I need real TLS certificates like I discussed in https://github.com/WebOfTrust/keripy/pull/570 

I could pull in test certificates like Sam did in the hio library here: https://github.com/ioflo/hio/blob/master/tests/core/tcp/test_tcp.py#L657

If you'd like me to do that let me know and I will. Otherwise the overall test coverage appears to be increasing, yet the patch coverage is still around 50%.

src/keria/app/agenting.py
kentbull: I was able to use the changes from #91 for a fully HTTPS KERIA. I haven't yet tried switching my witness over to HTTPS. That's my next task.

kentbull: After switching my witness over to HTTPS I was able to get a KERIA HTTPS agent server and a KERIpy HTTPS witness server to work using the code in this PR as well as Rodolfo's changes from #91.

pfeairheller: You still have not address my change request.  That change needs to be made before this PR can be accepted.

kentbull: Do you want to merge #103 and have me re-base this on development or should I just pull those changes into this branch?

kentbull: Ok, I'll pull just the changes I need into this PR.

rodolfomiranda:
seems that the codecov reports that is shown at the top of the PR gives different results that the one used to enforce the checks.

src/keria/app/agenting.py
kentbull: Those changes are now committed to this PR.

kentbull: nevermind, tests are breaking. I'll fix it.

rodolfomiranda:
I close PR #103 in favor of this one that includes all the changes

kentbull:
Is anything more needed for this PR to be merged?

PR #103, opened 2023-09-13
Support for https endpoints

rodolfomiranda:
This PR replace PR #91 due branch inconsistencies after moving from main to development.

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -899,8 +899,8 @@ class OobiResourceEnd:
                     raise falcon.HTTPNotFound(description=f"unable to query witness {wit}, no http endpoint")
 
                 url = urls[kering.Schemes.http] if kering.Schemes.http in urls else urls[kering.Schemes.https]
-                up = urlparse(urls[kering.Schemes.http])
-                oobis.append(f"{url}oobi/{hab.pre}/witness/{wit}")
+                up = urlparse(url)
+                oobis.append(urljoin(up.geturl(),f"/oobi/{hab.pre}/witness/{wit}"))
             res["oobis"] = oobis
         elif role in (kering.Roles.controller,):  # Fetch any controller URL OOBIs
             oobis = []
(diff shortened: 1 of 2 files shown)


src/keria/app/agenting.py
```diff
@@ -889,31 +888,34 @@ def on_get(req, rep, alias):
         if role in (kering.Roles.witness,):  # Fetch URL OOBIs for all witnesses
             oobis = []
             for wit in hab.kever.wits:
-                urls = hab.fetchUrls(eid=wit, scheme=kering.Schemes.http)
+                urls = hab.fetchUrls(eid=wit, scheme=kering.Schemes.http) or hab.fetchUrls(eid=wit, scheme=kering.Schemes.https)
```
m00sey, line 891: should add a test to cover this branch in logic

m00sey (changes requested):
tests

kentbull:
The main difference between these two PRs was the usage of `urljoin` to combine the parsed URL `up` with the rest of the OOBI URL string. I added that to #103 and we can merge that.

src/keria/app/agenting.py
kentbull: Agreed. We don't have coverage of that branch. In order to fully test that branch it seems we need to bring in locally self signed certificates for proper TLS. I could bring that in from HIO like I mentioned. Would you still prefer mocking in this case? We'd just need a hab with TLS urls. I'm not familiar with the test fixtures though I suppose there is a Hab test fixture pretty close to what we need.

kentbull: Yeah, looks like we'll need a new `helpers.openKeria()` test fixture that returns an agency started up with TLS config. Want me to make that for this PR?

rodolfomiranda:
I'm closing this PR since all the changes were included in PR #102

rodolfomiranda:
I'm closing this PR since all the changes were included in PR #102

PR #104, opened 2023-09-14
split Dockerfile build into two stages

kentbull:
One stage for building, one for the runtime. This enables near-instantaneous builds of source only changes

### DIFF
--- a/images/keria.dockerfile
+++ b/images/keria.dockerfile
@@ -1,31 +1,55 @@
+# Builder stage
+FROM python:3.10.13-alpine3.18 as builder
 
-FROM python:3.10.4-alpine3.16
+# Install compilation dependencies
+RUN apk --no-cache add \
+    bash \
+    alpine-sdk \
+    libffi-dev \
+    libsodium \
+    libsodium-dev
 
-RUN apk update
-RUN apk add bash
 SHELL ["/bin/bash", "-c"]
 
-RUN apk add alpine-sdk
-RUN apk add libffi-dev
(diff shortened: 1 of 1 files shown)


lenkan:
@kentbull This is great and speeds up the testing flow a great deal. Do you have any plans to merge this?

kentbull:
Though I don't have merge access it would be nice if this was merged.

lenkan:
I am using a different structure than this and have some thoughts... Do you need to use venv and a multi stage build to achieve this? I suggest keeping the single stage build and only separate the dependency and source with layers.

I suggest to use something like this:

```Dockerfile
FROM python:3.10.4-alpine3.16

RUN apk update && apk add --no-cache alpine-sdk libffi-dev libsodium libsodium-dev bash && curl https://sh.rustup.rs -sSf | bash -s -- -y

WORKDIR /keria
COPY requirements.txt setup.py /keria/
RUN mkdir /keria/src
RUN source "$HOME/.cargo/env" && pip install -r requirements.txt

COPY src/ src/
COPY scripts/keri/ scripts/keri/

ENTRYPOINT [ "keria" ]
CMD ["keria", "start",  "--config-file", "demo-witness-oobis", "--config-dir", "./scripts"]
```

Arguably, the config with hard coded witness oobis should be left out and provided by whoever starts the image. But that could wait.

pfeairheller (approved):

pfeairheller:
The codecov failures are not accurate here, going to merge.

PR #128, opened 2023-11-16
remove habname from exchange query

### LINKED ISSUE #127: How to get recipient of a notification message
Currently, on the holder/recipient side of a notification, the agent controller receives a notification containing this information:

```
{
  "i": "ABb0jd9TwszwDj_0jqTVFxkxXyqD10o-wluSoz4Z88x6", // Notification id
  "dt": "2023-11-16T13:28:40.988744",
  "r": false,
  "a": {
    "r": "/exn/ipex/grant",
    "d": "EMEtibwVhiwOTUVCW7S3eGribfJexxWFWL1saBUH7LbW", // EXN said
    "m": ""
  }
}
```

If the holder agent has more than one AID, is there a way to find which AID this message is intended for? I could walk through all AIDs and check GET /identifiers/{name}/exchanges/{said} for each, but that does not feel very elegant. Another option would be to filter the notifications by AID before showing them to the user, but currently that is not possible either from what I can see?

Looking more closely at the exchanges endpoints... It seems like neither endpoint to get exchange messages use the name parameter to filter out message for a specific hab:

- POST /identifiers/{name}/exchanges/query
- GET /identifiers/{name}/exchanges/{said}

They only do this check, then continuous to get the exchange messages, not using the `hab` or `name` variable.

```
        # Get the hab
        hab = agent.hby.habByName(name)
        if hab is None:
            raise falcon.HTTPNotFound(description=f"alias={name} is not a valid reference to an identifier")
```

See https://github.com/WebOfTrust/keria/blob/development/src/keria/peer/exchanging.py#L95-L137 and https://github.com/WebOfTrust/keria/blob/development/src/keria/peer/exchanging.py#L154-L169.

This make me think the endpoints could just be:

- POST /exchanges/query
- GET /exchanges/{said}

If so, I can just get the exchange message specified in the notification above and get the recipient AID from there.

lenkan:
Suggestion to resolve #127 

See issue for rationale.

### DIFF
--- a/src/keria/peer/exchanging.py
+++ b/src/keria/peer/exchanging.py
@@ -18,10 +18,10 @@ def loadEnds(app):
     app.add_route("/identifiers/{name}/exchanges", exnColEnd)
 
     exnColEnd = ExchangeQueryCollectionEnd()
-    app.add_route("/identifiers/{name}/exchanges/query", exnColEnd)
+    app.add_route("/exchanges/query", exnColEnd)
 
     exnResEnd = ExchangeResourceEnd()
-    app.add_route("/identifiers/{name}/exchanges/{said}", exnResEnd)
+    app.add_route("/exchanges/{said}", exnResEnd)
 
 
 class ExchangeCollectionEnd:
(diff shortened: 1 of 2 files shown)


lenkan:
We could also consider changing the POST /identifiers/{name}/exchanges to POST /exchanges and put the `name` parameter as part of the body. That would make it more coherent, but probably has more implications than I know of.

lenkan:
@pfeairheller Have you had a chance to look at this?

pfeairheller:
> @pfeairheller Have you had a chance to look at this?

Yes, just waited until this morning when I had a chance to update the SigPy client before merging.

PR #143, opened 2023-12-06
Fix ipex/grant endpoint

pfeairheller:
This PR fixes the ipex/grant endpoint to properly handle attachments.  There was a "disagreement" between clients and KERIA about what was being sent.


lenkan:
Great! I had the same issue and fixed it locally. But I won't submit a PR then.

PR #144, opened 2023-12-07
Simple healthcheck

kentbull:
Adds a simple healthcheck endpoint at each of the ports that is allowed through the SignifyTS header check. 

This is not a fully involved healthcheck that has useful statistics on the Agency, Habery, or Habs. That would be cool and will happen in a future evolution of the healthcheck. This endpoint is just to support a simple runtime status reporting that the application is running.

This is useful for Kubernetes readinessProbes and livenessProbes.


pfeairheller (changes requested):
I don't like exposing another un-authed endpoint for Agents.  In addition I don't think its needed as you can just use the get on state to see if an agent is awake.  Please remove this from the agent port and we can merge it just for Boot.

src/keria/app/agenting.py
```diff
@@ -774,6 +776,8 @@ def loadEnds(app):
     queryEnd = QueryCollectionEnd()
     app.add_route("/queries", queryEnd)
 
+    app.add_route("/health", HealthEnd())
```
pfeairheller, line 779: Please remove this route registration

src/keria/app/agenting.py
```diff
@@ -114,6 +115,7 @@ def setup(name, bran, adminPort, bootPort, base='', httpPort=None, configFile=No
 
         specEnd = AgentSpecResource(app=app, title='KERIA Interactive Web Interface API')
         specEnd.addRoutes(happ)
+        happ.add_route("/health", HealthEnd())
```
pfeairheller, line 118: Please remove this route registation

pfeairheller (changes requested):
Sorry for the delay getting back to this, but could you remove the route registrations.

src/keria/app/agenting.py
kentbull: done

src/keria/app/agenting.py
kentbull: done

kentbull:
Made all changes and then rebased on latest development, which I should'nt have. Now there's a failing test, `test_identifier_collection_end` in test_aiding` to unrelated functionality, when I run `pytest` locally.

Error message:
```python
            # Test Group Multisig
            keys = [signer0.verfer.qb64, p1.kever.verfers[0].qb64, p2.kever.verfers[0].qb64, ]
>           ndigs = [diger0.qb64, p1.kever.ndigers[0].qb64, p2.kever.ndigers[0].qb64]
E           AttributeError: 'Kever' object has no attribute 'ndigers'
```

m00sey (approved):

pfeairheller:
> Made all changes and then rebased on latest development, which I shouldn't have. Now there's a failing test, `test_identifier_collection_end` in test_aiding`to unrelated functionality, when I run`pytest` locally.
> 
> Error message:
> 
> ```python
>             # Test Group Multisig
>             keys = [signer0.verfer.qb64, p1.kever.verfers[0].qb64, p2.kever.verfers[0].qb64, ]
> >           ndigs = [diger0.qb64, p1.kever.ndigers[0].qb64, p2.kever.ndigers[0].qb64]
> E           AttributeError: 'Kever' object has no attribute 'ndigers'
> ```

You need to update your local environment to the latest KERIpy development branch

pfeairheller (approved):

PR #154, opened 2023-12-19
Update references to digers to be ndigers to match change in KERIpy.

pfeairheller:
This PR attempts to fix KERIA to match recent KERIpy changes.


m00sey (approved):
🥔

PR #181, opened 2024-02-02
Check agency db on cold start

kentbull:
On a restart of the KERIA service the agency has an empty list of agents and thus always tries to create an agent, which causes a 500. This manually checks the agency DB if the self.agency.agents is empty which fixes the problem.

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -818,7 +818,7 @@ class BootEnd:
 
         caid = icp.pre
 
-        if caid in self.agency.agents:
+        if self.agency.get(caid=caid) is not None:
             raise falcon.HTTPBadRequest(title="agent already exists",
                                         description=f"agent for controller {caid} already exists")
 


src/keria/app/agenting.py
```diff
@@ -818,7 +818,7 @@ def on_post(self, req, rep):
 
         caid = icp.pre
 
-        if caid in self.agency.agents:
+        if caid in self.agency.agents or self.agency.get(caid=caid) is not None:
```
pfeairheller, line 821: Instead of the `or`, just call self.agency.get(caid=caid) as the sole if condition.  That method checks the `in` anyway.

pfeairheller (changes requested):

src/keria/app/agenting.py
kentbull: done.

pfeairheller (approved):

PR #196, opened 2024-02-21
fixes for latest keripy development branch

m00sey:
Updates to support renaming in development keripy.


m00sey:
Merging this will require moving to latest keripy:development, currently we point to 1.1.0

Issue #193, opened 2024-02-13
IPEX apply, offer, agree implementation


iFergal:
I would like to get these added to KERIA and Signify-TS as it's something my team needs. Happy to give it a try myself, perhaps `apply` -> `offer` -> `agree` for a normal AID to begin with (multisig, `spurn` later).

**Understanding the existing impl:**
I've taken a look at the existing `grant` and `apply` messages in KERIA. The messages are parsed, pushed onto the `exchanging` deck to be sent to recipients and then pushed to either the `grants` or `admits` deck in `agent.ts`. From what I understand: `Granter` will send the other necessary material such as registry events, and `Admitter` will parse the ACDC, issuance and anchoring events attached to the admit message.

Q: `Admitter` will actually first check to see if there's a grant listed (otherwise invalid) - in case the grant is missing, or there is an issue parsing the attachments (e.g. the ACDC itself) hasn't the admit message already been sent to other parties using `exchanging` in `ipexing.py`? Curious if we should should check the grant is listed before sending the message onwards.

**New messages:**
`apply` and `agree` messages seem straight forward - do I need to do anything more other than sending the message to recipients in KERIA?

For `offer` I need a metadata ACDC, not sure if there's anything already available for this. Looking at the IPEX tests in `keripy`, it seems to just use a normal ACDC. Though it shouldn't be too hard to strip it down to a metadata ACDC and re-calculate the SAID.

iFergal:
@pfeairheller Have some questions re `Admitter` flow and the metadata ACDC in the description above if you have time to answer. Thanks :)

Just going to start looking at this today/this week - have been stuck on other things.

lenkan:
@iFergal can this be closed now?

PR #198, opened 2024-02-27
feat: IPEX apply, offer, agree endpoints

iFergal:
IPEX apply, offer, agree endpoints for non multi-sig AIDs. I can add multi-sig later if that's OK.

From https://github.com/WebOfTrust/keria/issues/193.

### DIFF
--- a/src/keria/app/ipexing.py
+++ b/src/keria/app/ipexing.py
@@ -20,13 +20,19 @@ def loadEnds(app):
     app.add_route("/identifiers/{name}/ipex/admit", admitColEnd)
     grantColEnd = IpexGrantCollectionEnd()
     app.add_route("/identifiers/{name}/ipex/grant", grantColEnd)
+    applyColEnd = IpexApplyCollectionEnd()
+    app.add_route("/identifiers/{name}/ipex/apply", applyColEnd)
+    offerColEnd = IpexOfferCollectionEnd()
+    app.add_route("/identifiers/{name}/ipex/offer", offerColEnd)
+    agreeColEnd = IpexAgreeCollectionEnd()
+    app.add_route("/identifiers/{name}/ipex/agree", agreeColEnd)
 
 
 class IpexAdmitCollectionEnd:
 
     @staticmethod
     def on_post(req, rep, name):
-        """  Registries GET endpoint
+        """ IPEX Admit POST endpoint
 
         Parameters:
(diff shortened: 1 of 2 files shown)


src/keria/app/ipexing.py
```diff
@@ -266,3 +272,203 @@ def sendMultisigExn(agent, hab, ked, sigs, atc, rec):
...
+        # Now create the stream to send, need the signer seal
+        kever = hab.kever
+        seal = eventing.SealEvent(i=hab.pre, s="{:x}".format(kever.lastEst.s), d=kever.lastEst.d)
+
+        ims = eventing.messagize(serder=serder, sigers=sigers, seal=seal)
+
+        # make a copy and parse
+        agent.hby.psr.parseOne(ims=bytearray(ims))
+
+        agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=rec, topic='credential'))
```
iFergal, line 473: I'm unsure if we are supposed to do something similar to admit (`Admitter`) where we now store the ACDC attachment that was offered, or if just the message is OK.

tests/app/test_ipexing.py
```diff
@@ -919,3 +925,314 @@ def test_granter(helpers):
...
+        pre = aid['i']
+        assert pre == "EHgwVwQT15OJvilVvW57HE4w0-GPs_Stj2OFoAHZSysY"
+
+        salt2 = b'0123456789abcdeg'
+        op = helpers.createAid(client, "recp", salt2)
+        aid1 = op["response"]
+        pre1 = aid1['i']
+        assert pre1 == "EFnYGvF_ENKJ_4PGsWsvfd_R6m5cN-3KYsz_0mAuNpCm"
+
+        # This should be a metadata ACDC in reality
```
iFergal, line 1034: Question here... if a metadata ACDC exists for making commitments to the metadata of a private ACDC - does this mean for a public ACDC you offer the full public ACDC or still the metadata ACDC?

src/keria/app/ipexing.py
pfeairheller: Just the message is OK here.

tests/app/test_ipexing.py
pfeairheller: I think it has to be the full public ACDC

pfeairheller (approved):

PR #221, opened 2024-04-04
Refactor Sealer to Anchorer

SmithSamuelM:


SmithSamuelM:
@pfeairheller I noticed that when I first checked out the development branch of Keria from WoT that not all the unit tests were successful.  So I am just leaving the pull request here. I don't think my changes broke anymore tests than before.

pfeairheller:
The merge of my smids and rmids change on KERIpy cause the problems because you did this before I could merge the matching PR here

I've now done so, if you rebase on development and repush you should be good to go

PR #224, opened 2024-04-04
Updates to match the changes in KERIpy for tracking multisig participants.

pfeairheller:
This PR includes:

- Changes to account for new tracking of SMIDs and RMIDs in KERIpy
- Changes to account for the new database tables for tracking Habs by AID and name in KERIpy.
- Test fixes.

### DIFF
--- a/.github/workflows/python-app-ci.yml
+++ b/.github/workflows/python-app-ci.yml
@@ -21,10 +21,10 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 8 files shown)


SmithSamuelM:
I am getting a failing test on my local copy of KERIA after I merged in this commit.
It is test_aiding.py  test_oobi_ends()

So maybe we can figure out what its expecting  that I don't have.

Also I set up my venv so that the setup.py dependency works for keri without using the URL but maybe there is something else I need to understand.

PR #234, opened 2024-04-26
Return 400 when Serder validation fails

rodolfomiranda:
This PR is related to #201

PR #237, opened 2024-04-30
Fix references to TraitCodex after move

lenkan:
TraitCodex was moved in https://github.com/WebOfTrust/keripy/commit/a1c4d26895acb6117b96e36ae14aaedfc233c4f6

Pinned the macos version as in #234.

### DIFF
--- a/.github/workflows/python-app-ci.yml
+++ b/.github/workflows/python-app-ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ macos-latest, ubuntu-latest ]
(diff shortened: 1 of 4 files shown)


pfeairheller:
Because this PR was showing at not built, I thought you hadn't made the MacOS change.  But you did and hit the same problem I had, it somehow still thinks we need a macos-latest.

I made these exact changes in my PR and merged it so I'm closing this one.  Hopefully the main build will work now.

lenkan:
@pfeairheller It looks like it is the branch protection rules that requires the build for macos-latest to pass before merging. Should be a quick change in 

Settings => Branches => edit main => Require status checks to pass before merging => Replace macos-latest with macos-13

PR #242, opened 2024-05-14
build: pin exact keripy version in setup.py

### LINKED ISSUE #163: Locking keripy dependency
Opening a new issue for this more generally but #152 already mentions this. Would it be OK if we pin the githash of keripy used? e.g. `keri @ git+https://git@github.com/WebOfTrust/keripy.git@2d530601d3463d820229de0676cda519523b73ca`

Right now we are lacking determinism for building KERIA - building older versions typically break, and even the latest can break if it falls slightly behind keripy.

This in general causes problems with deployment to environments with multiple team members, git bisect, CICD determinism etc.

iFergal:
To close #163 - `>=` is pulling `1.2.0-dev2` for me in a fresh build which is breaking. In absence of a lockfile, `==` brings better determinism in our builds.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ setup(
     python_requires='>=3.12.2',
     install_requires=[
         'hio>=0.6.12',
-        'keri>=1.2.0-dev0',
+        'keri==1.2.0-dev0',
         'mnemonic>=0.20',
         'multicommand>=1.0.0',
         'falcon>=3.1.3',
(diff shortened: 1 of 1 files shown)


pfeairheller:
Looks like you at least need dev1

iFergal:
Yeah, `0.2.0-dev0` tag required `dev0` which is where I saw the issue originally, forgot main would be different.

Has passed now with `dev2` which is what the last PR build on `main` used so I just matched that.

pfeairheller (approved):

PR #247, opened 2024-05-23
added approval endpoint for delegator after they anchor seal in their KEL

2byrds:
Additions based on Phil's guidance and using his current PR

### DIFF
--- a/src/keria/app/delegating.py
+++ b/src/keria/app/delegating.py
@@ -109,9 +109,7 @@ class Anchorer(doing.DoDoer):
 
     def processUnanchoredEscrow(self):
         """
-        Process escrow of partially signed multisig group KEL events.  Message
-        processing will send this local controllers signature to all other participants
-        then this escrow waits for signatures from all other participants
+        Process escrow of unacnchored events that have been delegated and are waiting for delegator anchor or approval.
 
         """
         for (pre, said), serder in self.hby.db.dune.getItemIter():  # group partial witness escrow
--- a/src/keria/app/delegating.py
+++ b/src/keria/app/delegating.py
@@ -109,7 +109,7 @@ class Anchorer(doing.DoDoer):
 
     def processUnanchoredEscrow(self):
         """
-        Process escrow of unacnchored events that have been delegated and are waiting for delegator anchor or approval.
+        Process escrow of unanchored events that have been delegated and are waiting for delegator anchor or approval.
 
         """
         for (pre, said), serder in self.hby.db.dune.getItemIter():  # group partial witness escrow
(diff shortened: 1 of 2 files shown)


2byrds:
> * How does it work when the delegator is a multisig? Do all participants have to do the same thing?

Great questions @lenkan i'll need to add tests for the multi-sig case to be sure. The delegatee 'dip' event and the 'ixn' events will be thresholded, the multi-sig AID won't complete until enough signatures are provided.

> * Does the second POST with the approveDelegation field actually append an event? If not, I think this should move to a separate endpoint to avoid confusion. To me POST /identifiers/name/events says that you will append an event to the KEL.

Without the approval, the delegator AID won't process the 'dip' of the delegatee. It will sit in escrow until the approval is given. It is meant to be similar to the 'ixn' anchor being needed, but this is an additional check in order to protect the delegator AID from accidentally anchoring a seal that triggers an approval. So the delegation isn't complete until the 'dip' is finally processed by the delegator when the approval is sent, but that still depends on all the signature thresholds being met. Since it is a two step-process the correct functionality is that a threshold of approvals is met, I believe. I'll verify with tests if a threshold of approvals is needed for the multi-sig case. Thoughts?

src/keria/app/delegating.py
```diff
@@ -109,9 +109,7 @@ def processEscrows(self):
 
     def processUnanchoredEscrow(self):
         """
-        Process escrow of partially signed multisig group KEL events.  Message
-        processing will send this local controllers signature to all other participants
-        then this escrow waits for signatures from all other participants
+        Process escrow of unacnchored events that have been delegated and are waiting for delegator anchor or approval.
```
lenkan, line 112:
```suggestion
        Process escrow of unanchored events that have been delegated and are waiting for delegator anchor or approval.
```

lenkan:
I find the API little confusing.

What is the process?

1. Delegatee create delegated inception event `dip`
1. Delegator receives the inception event details out of band
1. Delegator create interaction event with delegation information in anchor data.
1. Delegator approves delegation by posting the same interaction event again?

What is the risk with having step 4 processed automatically?

If it is not possible to do automatically, can we add a flag to step 3 so it becomes

```
POST /identifiers/name/events
{ 
   ixn: [ anchor data ], 
   approveDelegation: true 
}
```

Another option if we need to explicitly call the server again, can we move it to another endpoint? What is it that is being created or modified here? Is it a "pending delegation request"?

Note: I am not saying any of my suggestion are objectively better, just trying to trigger discussion about how we design the HTTP API of KERIA so that we can improve it as we move forward.

2byrds:
Shifting over to a new PR, but will recap the needed changes based on @lenkan suggestions #248

PR #250, opened 2024-06-04
Feat/delegation approval endpoint

2byrds:
A fresh PR based on work and feedback from https://github.com/WebOfTrust/keria/pull/248

1. I will change the keri dependency to the 1.2.0-dev5 tag once it is available.
1. Adds delegation logic to delegating.py and tests in test_delegation.py
1. Adds "/identifiers/{name}/delegation" end point per review comments from previous PR.
1. Passes IdentifierResourceEnd, similar to how credentialing.py, per previous PR review comment.
1. Only addresses the interaction case. A separate issue can be generated for handling/testing the rotation case.
1. Extends long running 'delegation' operation to handle delegator approval case.
1. Integration tested with updates to signify-ts which adds support for the delegation endpoint.


src/keria/app/delegating.py
```diff
@@ -172,4 +178,71 @@ def processPartialWitnessEscrow(self):
...
+            if pre == teepre:
+                seqner = coring.Seqner(sn=serder.sn)
+                couple = seqner.qb64b + serder.saidb
+                dgkey = dbing.dgKey(coring.Saider(qb64=teepre).qb64b, coring.Saider(qb64=teesaid).qb64b)
+                # the dip event should have been received from the delegatee via a postman call
+                # and will be sitting in the delegator escrows (hence the hab.db.delegables above)
+                # adding the authorize event seal will allow the dip to be processed
+                # and added to the delegator kever
+                hab.db.setAes(dgkey, couple)  # authorizer event seal (delegator/issuer)
+                return teepre
```
pfeairheller, line 246: I'm still not seeing where you reprocess the delegate's event from delegables and then delete it from delegables.

pfeairheller (changes requested):
There is just one thing missing I think unless I missed it somewhere.  Address or answer that and we'll be good to merge.

src/keria/app/delegating.py
2byrds: Thank you @pfeairheller ! based on my debugging, I see the order of operations as:
- [after the delegatee completes witness receipts of the dip, it is sent to the delegator](https://github.com/WebOfTrust/keripy/blob/38cc2891286afce6c7426b0bd14abca287f229fc/src/keri/app/delegating.py#L186)
- [this dip event is escrowed by the delegator](https://github.com/WebOfTrust/keripy/blob/38cc2891286afce6c7426b0bd14abca287f229fc/src/keri/core/eventing.py#L2949)
- The delegator submits the anchor ixn event, but that alone doesn't allow the escrowed dip to be processed. That is because the processEscrowDelegables continues to [fail this check](https://github.com/WebOfTrust/keripy/blob/38cc2891286afce6c7426b0bd14abca287f229fc/src/keri/core/eventing.py#L5800)
- We added an additional step of approving delegation which only calls setAes with the ixn anchor info.
- The next time the processEscrowDelegables runs, the Aes check passes which [allows the delegator to process the dip successfully](https://github.com/WebOfTrust/keripy/blob/38cc2891286afce6c7426b0bd14abca287f229fc/src/keri/core/eventing.py#L5806)
- And then for the [delegables to be removed](https://github.com/WebOfTrust/keripy/blob/38cc2891286afce6c7426b0bd14abca287f229fc/src/keri/core/eventing.py#L5830)
- This results in the delegator having the delegatee in its kevers and for the delegation to be complete

pfeairheller: Now I see the disconnect... processEscrowDelegables is never called anywhere.  Here or KERIpy.

pfeairheller (approved):

pfeairheller:
You need to rebase with main to resolve the conflict.

PR #256, opened 2024-06-19
Increase Escrower tock to avoid overly busy CPU during escrow processing

rubelhassan:
Setting tock to 1.0 second runs escrower loop 30s delay and delegator approval received after 30 seconds

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -5,7 +5,6 @@ keria.app.agenting module
 
 """
 import json
-import datetime
 import os
 from dataclasses import asdict
 from urllib.parse import urlparse, urljoin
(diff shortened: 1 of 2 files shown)


src/keria/app/agenting.py
```diff
@@ -745,23 +746,24 @@ def __init__(self, kvy, rgy, rvy, tvy, exc, vry, registrar, credentialer):
...
-        self.rvy.processEscrowReply()
-        if self.tvy is not None:
-            self.tvy.processEscrows()
-        self.exc.processEscrow()
-        self.vry.processEscrows()
-        self.registrar.processEscrows()
-        self.credentialer.processEscrows()
-
-        return False
+        while True:
```
pfeairheller, line 755: This change is not needed.  Once a Doer has a self.tock value, the parent Doist/DoDoer will use that so you don't need to turn recur into a generator just to yield the tock.

pfeairheller (changes requested):

src/keria/app/agenting.py
rubelhassan: I didn't notice that than you for your feedback. I've updated the PR.

pfeairheller (approved):

PR #262, opened 2024-07-01
Removing AID delete endpoint

rodolfomiranda:
This PR removes the `delete` AID endpoint I created before. The implementation wasn't correct since it only removes the `hab` generation inconsistencies in the database if the AID also had registries. Rethinking the concept, I consider that deleting an AID in not what we need, but instead an implementation of the `abandon` function that basically rotates the identifier to null keys so it can no longer be used to sign nor add new events in into the KEL.
The `abandon` function can be created in the client side since its a double rotation to null keys.


lenkan:
@rodolfomiranda What about adding a check and return Bad Request if the AID to be deleted has a registry?

rodolfomiranda:
we can delete the registry as well, but I think it's better to keep KEL and TEL in DBs for an AID. The only benefit of deleting is to prune database. However the saving in size is minimal to be of a concern.
That's why I'd prefer to abandoned it, it's more `keri` friendly . Detecting an abandoned AID can be simple, we can even put a filter in the `identifiers.list()`

lenkan:
> we can delete the registry as well, but I think it's better to keep KEL and TEL in DBs for an AID. The only benefit of deleting is to prune database. However the saving in size is minimal to be of a concern. That's why I'd prefer to abandoned it, it's more `keri` friendly . Detecting an abandoned AID can be simple, we can even put a filter in the `identifiers.list()`

I also think maybe they can serve different purposes. For certain environments, like demo or testing, it can sometimes be useful to be able to delete them altogether. However, I do not really have a strong opinion on it. The proposed `abandon` feature would certainly work for those use cases as well. 👍

rodolfomiranda:
I opened this PR because we allowed one user to delete the AID and it worked fine until he needed to reconnect with KERIA. At that point,  keria iterates through all registries for that user and crash when it hits the one with the deleted `hab`. We had to create a hot patch to recover from that situation. I'm worried that it will also happen with other tables/db since the dependencies between them are hard to track.

We main need a keria maintenance interface.

pfeairheller (approved):

Issue #217, opened 2024-03-28
Support AID instead of or in addition to name in URLs that require it


pfeairheller:
More details to follow

daviddm:
Also related https://github.com/WebOfTrust/keria/issues/202

lenkan:
@pfeairheller I think this can be closed after #269 

cc @2byrds

Issue #202, opened 2024-03-01
Provide a way to get identifier by prefix


lenkan:
It would be convenient to be able to do `GET /identifiers/{prefix}` instead of `GET /identifiers/{name}`. One use case is when you are reading a notification or exchange message. In that case, you only have the prefix of the intended recipient AID. So currently you have to first need an API call to get all identifiers, then pick the identifier you need to work with.

```
exchange = { ... a: { i: "prefix-1" } }
identifiers = listIdentifiers() // API Request to list all identifiers
identifier = identifiers.find(id => id.prefix === exchange.a.i) // Pick the identifier you needed
```

If we had a way to get identifier by prefix, we could do

```
exchange = { ... a: { i: "prefix-1" } }
identifier = getIdentifier(exchange.a.i)
```

# Notes

Currently, you can only get an identifier by their name through the API. This is convenient for manually constructing URLs or when using the `kli`. When code is constructing the URL, I think using the prefix is more useful.

2byrds:
From our dev meeting this issue could be closed in favor https://github.com/WebOfTrust/keria/issues/217

lenkan:
Fixed by #269

PR #264, opened 2024-07-02
Allowing prefixes in identifiers endpoint

rodolfomiranda:
This PR is an implementation of #217 and #202, fixing the problem of aliases with special characters .
The changes allow endpoints of the form `/identifiers/{name}` to accept an alias or a prefix in the name path. If the value is valid for both prefix and alias, the prefix is prioritized.
Basically, the PR replace:
`hab = agent.hby.habByName(name)`
with
`hab = agent.hby.habs[name] if name in agent.hby.habs else agent.hby.habByName(name)`


pfeairheller (changes requested):
Could you please also update the doc strings here for the description of the name parameter since these are used in the OpenAPI document generation?

rodolfomiranda:
I screw up the PR when rebasing. It picked other commits. I created a new clean one in #269 and I'm closing this.

Issue #266, opened 2024-07-02
Add features of filtering and sorting in KERIA notifications


rodolfomiranda:
Notification endpoint in KERIA already provide pagination with the `range` headers. We are looking to enhance the endpoint with filters by notification type (`note.a.r`) and read status (`note.r`).

PR #267, opened 2024-07-05
Add filters in notifications endpoint

rodolfomiranda:
This PR address issue #266 . It adds to `NotificationCollectionEnd` three **optional** query parameters
- `read` to filter by the `read` property of the note (true or false).
- `route` to filter by the route `r` property in the attributes section (`a`) if exists.
- `order` to order the results by `datetime` in `asc` or `desc` order.

### DIFF
--- a/src/keria/app/notifying.py
+++ b/src/keria/app/notifying.py
@@ -49,6 +49,10 @@ class NotificationCollectionEnd:
         """
         agent = req.context.agent
 
+        read = req.get_param_as_bool("read")
+        route = req.get_param("route")
+        order = req.get_param("order")
+
         rng = req.get_header("Range")
         if rng is None:
             rep.status = falcon.HTTP_200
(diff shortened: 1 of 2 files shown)


lenkan:
Very helpful feature!

Question: I am not familiar with the notifier implementation, but this looks like it would load all notes in to memory, then filter them? Would it make sense to instead implement the filtering in [keripy](https://github.com/WebOfTrust/keripy/blob/6b6970be92d851f94e58b7a9f4225a81582d2235/src/keri/app/notifying.py#L327), utilizing the iterator to the need to load everything into memory?

rodolfomiranda:
That's a good point. The optimal way is to implement it in `keripy` but it needs to run the iterator also on the `countAll` function. I can move it to there easily, but I need the feedback from others specially @pfeairheller. If we are in agreement, I'll close this one and mimic it in keripy.

rodolfomiranda:
I created a PR in `KERIPY` (https://github.com/WebOfTrust/keripy/pull/816) that should be a more efficient way to implement the ordering and filtering.
I'll put this PR in draft until the keripy PR is being review. If it's merged, I'll adapt this one.

pfeairheller:
I've commented and requested changes in (https://github.com/WebOfTrust/keripy/pull/816).  We can readdress this issue when that is resolved.

iFergal:
@kentbull I think this PR can be closed given it's inactivity - it should be handled at a DB level rather than in memory in keripy, and the keripy comments were never resolved.

Issue #235, opened 2024-04-29
Provide a way to re-submit AID KEL to witnesses


lenkan:
If a witness is missing data, for example as a result of the issue described in https://github.com/WebOfTrust/keripy/issues/762, we need a way to re-submit the KEL of an AID to its witnesses. Similar to the `kli witness submit` command.

Quote from issue:

> Can this be repaired? KERIA probably not. From command line witness submit could catch it up. Possibly 'witness submit' in KERIA could be a workaround.

2byrds:
fixed by https://github.com/WebOfTrust/keria/pull/271

PR #268, opened 2024-07-08
updates for escrow fix, align with kerupy 1.1.17 and #235

m00sey:


2byrds (changes requested):
Please remove the 'Submitter' functionality/endpoint you have added. It is a partial addition that will conflict with the work i'm doing here https://github.com/2byrds/keria/tree/feat/submit_id

m00sey:
Ned to update keripy ref to pull in hio fix

PR #269, opened 2024-07-11
Allowing prefixes in identifiers endpoint (resubmitted PR)

rodolfomiranda:
This PR is resubmitted version of the canceled #264 
It's is an implementation of #217 and #202, fixing the problem of aliases with special characters .
The changes allow endpoints of the form `/identifiers/{name}` to accept an alias or a prefix in the name path. If the value is valid for both prefix and alias, the prefix is prioritized.
Basically, the PR replace:
`hab = agent.hby.habByName(name)`
with
`hab = agent.hby.habs[name] if name in agent.hby.habs else agent.hby.habByName(name)`


2byrds (changes requested):
@rodolfomiranda this looks like a nice addition. I see modifications to the test error response, is it also possible to add some access by prefix and access by name in each test to show that both functionalities work?

2byrds (approved):

Issue #257, opened 2024-06-26
Internal Server Error on connect after upgrade from 0.1.2 to 0.2.0-dev2 when agent contains group hab


lenkan:
I am having issues upgrading a keria instance from 0.1.2 to 0.2.0-dev2. When the instance contains an agent with a group hab, the controller cannot connect to it.

See reproduction here: https://github.com/lenkan/keria-upgrade-test. It should be possible to run it by cloning the repo and 

```
git clone git@github.com:lenkan/keria-upgrade-test.git
cd keria-upgrade-test
./run-test.sh
```

The error output is:

```
connect-1  | /app/scripts-connect/node_modules/signify-ts/src/keri/app/controller.ts:152
connect-1  |         if (state == null || state['ee']['s'] == 0) {
connect-1  |                                        ^
connect-1  |
connect-1  |
connect-1  | TypeError: Cannot read properties of undefined (reading 's')
connect-1  |     at Controller (/app/scripts-connect/node_modules/signify-ts/src/keri/app/controller.ts:152:40)
connect-1  |     at SignifyClient.connect (/app/scripts-connect/node_modules/signify-ts/src/keri/app/clienting.ts:133:27)
connect-1  |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
connect-1  |     at <anonymous> (/app/scripts-connect/src/main.ts:13:1)
connect-1  |
connect-1  | Node.js v20.12.2
connect-1  | npm ERR! Lifecycle script `start` failed with error:
connect-1  | npm ERR! Error: command failed
connect-1  | npm ERR!   in workspace: undefined
connect-1  | npm ERR!   at location: /app/scripts-connect
connect-1 exited with code 1
keria-1  | /keripy/src/keri/core/serdering.py:120: SyntaxWarning: invalid escape sequence '\A'
keria-1  |   """Design Notes:
keria-1  | The Agency is loaded and waiting for requests...
keria-1  | 2024-06-26 13:21:20 [FALCON] [ERROR] GET /agent/EEFCS78eJySGGx9Nl7r-eHRSDZ4xbisWwkqrwEWft-bG => Traceback (most recent call last):
keria-1  |   File "falcon/app.py", line 365, in falcon.app.App.__call__
keria-1  |   File "/usr/local/var/keria/src/keria/app/aiding.py", line 110, in on_get
keria-1  |     agent = self.agency.get(caid)
keria-1  |             ^^^^^^^^^^^^^^^^^^^^^
[...]

PR #272, opened 2024-07-18
feat: migrate command for all agents

iFergal:
This PR adds a `keria migrate` command that will fetch all of the agents from the agency DB and kick off the migrate doer used in `kli migrate`. Slightly weird to mimic a kli command from keria but it works, and means we will pick up changes to `kli migrate`.

I also pinned the keripy version again - this got unpinned by mistake it seems. But as mentioned before we need it for determinism of builds.

Here is a screenshot example of incepting 2-of-2 multi-sig on an old version, trying to connect without migrating and successfully migrating:
image

**Note:** `SignifyGroupHabs` are not correctly migrated on keripy - but this is separate for now and has been raised in #257 - I gave some details on why in that issue yesterday.

### DIFF
--- a/src/keria/app/cli/commands/migrate.py
+++ b/src/keria/app/cli/commands/migrate.py
@@ -1,0 +1,56 @@
+# -*- encoding: utf-8 -*-
+"""
+KERIA
+keria.cli.keria.commands module
+
+"""
+import argparse
+
+from hio.base import doing
+from keri.app.cli.commands.migrate import run
+from keria.db import basing
+
+
+def handler(args):
+    """
+    Migrate KERIA agents
+
+    Args:
+        args(Namespace): arguments object from command line
+    """
+    migrator = MigrateDoDoer(args)
+    return [migrator]
+
+
+parser = argparse.ArgumentParser(description='Runs outstanding migrations for all agents')
+parser.set_defaults(handler=handler,
+                    transferable=True)
+parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',
+                    required=False, default="")
+
+
+class MigrateDoDoer(doing.DoDoer):
+    def __init__(self, args):
+        self.args = args
+        self.adb = basing.AgencyBaser(name="TheAgency", base=args.base, reopen=True, temp=False)
+        doers = [doing.doify(self.migrateAgentsDo)]
+
+        super(MigrateDoDoer, self).__init__(doers=doers)
+
+    def migrateAgentsDo(self, tymth, tock=0.0):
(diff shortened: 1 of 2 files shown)


2byrds:
Looks good to me @iFergal. I suppose this is more of an administrative feature, hence the command line tool. Do you have any thoughts how we can test a command like this in case migrate changes?

2byrds:
Also @iFergal could you post a screenshot(s) showing the command being successfully used on this PR?

iFergal:
Thanks @2byrds I added a screenshot and note to the description.

Yeah based on recommendations from Phil, migrations should be manual, hence the cli.

Regarding testing - I think we would need a minimal agent and set the migration version to the lowest, run `keria migrate` and then the basic check could be that the version is up to date. I'm not sure if creating an agent would conflict with any schemas that are assumed to be not migrated already (so far, maybe not, but maybe in the future) - but I could look into it.

On the other hand it's probably unlikely that the `kli migrate` command will be subject to much change since it just opens the DB and calls `db.migrate()`.

iFergal:
Based on discussions yesterday and https://github.com/WebOfTrust/keripy/issues/824 more thought should be put into this.

If SignifyGroupHab and other files move to KERIA, we may need a separate set of migrations for KERIA on top of calling `db.migrate()` from keripy.

src/keria/app/cli/commands/migrate.py
```diff
@@ -0,0 +1,56 @@
...
+                    transferable=True)
+parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',
+                    required=False, default="")
+
+
+class MigrateDoDoer(doing.DoDoer):
+    def __init__(self, args):
+        self.args = args
+        self.adb = basing.AgencyBaser(name="TheAgency", base=args.base, reopen=True, temp=False)
+        doers = [doing.doify(self.migrateAgentsDo)]
```
kentbull, line 36: Let's get away from using `doing.doify` since Sam has cautioned against this. I am willing to show how to do this by refactoring the PR. You replace `doing.doify` with a Doer class that has the logic called from the `Doer.recur()` function.

I will write a HIO and KERIpy style guide to help communicate some of these ideas.

kentbull (approved):
Looks good enough to me. We could do a code style change as I mentioned, yet that's not critical.

iFergal:
@kentbull I hadn't had the time to revisit this since we had the discussion that parts of keripy need to be moved here, in which case this migration needs to be handled differently. I would be in favour of closing it until someone has capacity to handle it properly.

kentbull:
Okay, I'll close this for now.

PR #271, opened 2024-07-18
Feat/submit

2byrds:
Adds a KERIA endpoint /identifiers/${name}/submit to resubmit an identifier KEL to be receipted by it's witnesses.
Adds 'submit' unit tests
This feature is the KERIA equivalent to the keripy [witness 'submit' command line tool](https://github.com/WebOfTrust/keripy/blob/3febc54a463125a4cc3d86d694e57004134b2092/src/keri/app/cli/commands/witness/submit.py#L61)

PR #254, opened 2024-06-13
Fix/del rot smid rmid tests

2byrds:
Catches KERIA up with KERIpy changes AND provides all functionality necessary for SIGNIFY-TS integration tests to pass, especially multisig-vlei-issuance but also fixes for rotation, salty, and delegation

PR #273, opened 2024-07-19
Action to enforce python formatting

2byrds:
Introduce a github action to check python formatting for all PRs being submitted to main, in order to align formatting.

### DIFF
--- a/.github/workflows/python-formatting.yml
+++ b/.github/workflows/python-formatting.yml
@@ -26,15 +26,3 @@ jobs:
 
     - name: Run black
       run: black --check .
-
-    - name: Post comment on PR if formatting fails
-      if: failure()
-      uses: actions/github-script@v3
-      with:
-        script: |
-          github.rest.issues.createComment({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            body: `Formatting check failed. Please run \`black .\` to format your code. For more details, please check the 'Checks' tab of this PR.`
-          });
--- a/.github/workflows/python-formatting.yml
+++ b/.github/workflows/python-formatting.yml
@@ -30,3 +30,12 @@ jobs:
     - name: Run black to auto-format
       if: failure()
       run: black .
+
+    - name: Commit and push changes
+      if: failure()
+      run: |
+        git config --global user.name 'github-actions[bot]'
+        git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+        git add .
+        git commit -m 'Auto-format code with black'
+        git push
--- a/.github/workflows/python-formatting.yml
+++ b/.github/workflows/python-formatting.yml
@@ -27,15 +27,16 @@ jobs:
     - name: Run black
       run: black --check .
 
-    - name: Run black to auto-format
+    - name: Post comment on PR if formatting fails
       if: failure()
-      run: black .
-
-    - name: Commit and push changes
-      if: failure()
-      run: |
-        git config --global user.name 'github-actions[bot]'
-        git config --global user.email 'github-actions[bot]@users.noreply.github.com'
-        git add .
-        git commit -m 'Auto-format code with black'
-        git push
+      uses: actions/github-script@v3
+      with:
+        script: |
+          const { execSync } = require('child_process');
+          const result = execSync('black --check .', { encoding: 'utf-8' });
+          github.rest.issues.createComment({
+            issue_number: context.issue.number,
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            body: `Formatting check failed. Please run \`black .\` to format your code.\n\n\`\`\`\n${result}\n\`\`\``
+          });
--- a/.github/workflows/python-formatting.yml
+++ b/.github/workflows/python-formatting.yml
@@ -25,4 +25,4 @@ jobs:
       run: pip install autopep8
 
     - name: Run autopep8
-      run: autopep8 diff --exit-code .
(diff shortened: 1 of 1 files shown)


m00sey:
I think this is solving the wrong problem.

Rather than "reformatting" the entire repo, making future diffs unintelligible, I'd rather see a GitHub action that enforces the _current_ format, rather than switching formats entirely.

2byrds:
> I think this is solving the wrong problem.
> 
> Rather than "reformatting" the entire repo, making future diffs unintelligible, I'd rather see a GitHub action that enforces the _current_ format, rather than switching formats entirely.

@m00sey what is the current format(er)?
If the current format(er) is well known then we can apply a check for it. If it is not well known then if we can transition it to a well known format then it is easy for every PR diff to be well formatted, making a diff highlight only content changes (which is ideal)

2byrds:
> I think this is solving the wrong problem.
> 
> Rather than "reformatting" the entire repo, making future diffs unintelligible, I'd rather see a GitHub action that enforces the _current_ format, rather than switching formats entirely.

@m00sey I've updated the title, the action only checks the formatting now (enforcing well formatted PRs). It does not try to format it for you any longer. That was discussed on https://github.com/WebOfTrust/keria/pull/271 and I had updated it earlier today when I realized that auto-formatting was overkill.

m00sey:
> > I think this is solving the wrong problem.
> > Rather than "reformatting" the entire repo, making future diffs unintelligible, I'd rather see a GitHub action that enforces the _current_ format, rather than switching formats entirely.
> 
> @m00sey what is the current format(er)? If the current format(er) is well known then we can apply a check for it. If it is not well known then if we can transition it to a well known format then it is easy for every PR diff to be well formatted, making a diff highlight only content changes (which is ideal)

The current formatting checks are in the CI file:

https://github.com/WebOfTrust/keria/blob/c83904d62b908e5f030bbf7f957d397cd259035f/.github/workflows/python-app-ci.yml#L35

Anything proposed down this route should start here.

2byrds:
> > > I think this is solving the wrong problem.
> > > Rather than "reformatting" the entire repo, making future diffs unintelligible, I'd rather see a GitHub action that enforces the _current_ format, rather than switching formats entirely.
> > 
> > 
> > @m00sey what is the current format(er)? If the current format(er) is well known then we can apply a check for it. If it is not well known then if we can transition it to a well known format then it is easy for every PR diff to be well formatted, making a diff highlight only content changes (which is ideal)
> 
> The current formatting checks are in the CI file:
> 
> https://github.com/WebOfTrust/keria/blob/c83904d62b908e5f030bbf7f957d397cd259035f/.github/workflows/python-app-ci.yml#L35
> 
> Anything proposed down this route should start here.
> 
> E9: Syntax Errors - These errors occur when there is an issue with the syntax in the Python code, such as a missing colon, unmatched parentheses, or other fundamental syntax mistakes. F63: Problem with Import - This code indicates that there is a problem with an import statement, typically an issue with module-level import. F7: Errors related to function definitions - This includes issues such as missing arguments in function definitions or problems with lambda functions. F82: Unused local variable - This error occurs when a local variable is defined but not used anywhere in the function or code block.
> 
> Ignoring: E7: Syntax errors related to tokenization - This error indicates issues during the tokenization process, often related to unmatched parentheses or brackets. F841: Local variable assigned but not used - This occurs when a local variable is assigned a value but is not used in any operations within the code block. E301: Expected 1 blank line, found 0 - This error is raised when there is not a blank line after a function or class definition. E302: Expected 2 blank lines, found 1 - This error occurs when there should be two blank lines before a function or class definition, but there is only one. E303: Too many blank lines (3) - This error occurs when there are more than two consecutive blank lines in the code.
> 
> https://flake8.pycqa.org/en/latest/

The enabled rules wouldn't help enforce formatting issues for PRs. It is leveraging the syntax linting checks for flake8, and not providing formatting checks. To help distinguish between linters and formatters this is a decent quick read https://py-vscode.readthedocs.io/en/latest/files/linting.html

m00sey:
Okie - just trying to give you a jumping off point.

You managed to create PRs without mass reformatting previously - https://github.com/WebOfTrust/keria/pull/254/files

Maybe you should just go back to that workflow?

2byrds:
> Okie - just trying to give you a jumping off point.
> 
> You managed to create PRs without mass reformatting previously - https://github.com/WebOfTrust/keria/pull/254/files
> 
> Maybe you should just go back to that workflow?
> 
> Seems like that would expedite the merging process.

I'm not sure what you mean. We want to avoid formatting from polluting PRs now and in the future. If we agree on a formatter than that can be easily checked. I have updated this PR to use autopep8 python formatter since @pfeairheller noted he uses PEP in IntelliJ. You can see what the output would look like on the latest formatting check for this PR. I still have no preference of what formatter to use, only that we make formatting simple now and in the future.

.github/workflows/python-formatting.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Check out the repository
+      uses: actions/checkout@v2
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
```
tibor19, line 22: Given that the docker image uses python 3.12 wouldn't make more sense to use the same version explicitly?

2byrds: @tibor19 I think its good to be consistent with the other checks using 3.12.2, so i have updated this in the new PR https://github.com/WebOfTrust/keria/pull/280

PR #278, opened 2024-07-30
feat: multi-sig IPEX apply, offer, agree endpoints

iFergal:
Continuing #193 for multi-sig. No spurn yet.

### DIFF
--- a/src/keria/app/ipexing.py
+++ b/src/keria/app/ipexing.py
@@ -119,7 +119,7 @@ class IpexAdmitCollectionEnd:
 
         # Have to add the atc to the end... this will be Pathed signatures for embeds
         if not atc:
-            raise falcon.HTTPBadRequest(description=f"attachment missing for ACDC, unable to process request.")
+            raise falcon.HTTPBadRequest(description=f"attachment missing for multi-sig admit, unable to process request.")
 
         # use that data to create th Serder and Sigers for the exn
         serder = serdering.SerderKERI(sad=ked)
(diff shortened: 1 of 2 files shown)


src/keria/app/ipexing.py
```diff
@@ -144,6 +143,8 @@ def sendMultisigExn(agent, hab, ked, sigs, atc, rec):
         if grant is None:
             raise falcon.HTTPBadRequest(description=f"attempt to admit an invalid grant {admitked['p']}")
 
+        agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=rec, topic='credential'))
```
iFergal, line 146: I've moved this below the above checks for the exn message and link to grant etc - presume we don't want to send the `/multisig/exn` message to others until we have passed those checks

tests/app/test_ipexing.py
```diff
@@ -341,24 +363,32 @@ def test_ipex_grant(helpers, mockHelpingNowIso8601, seeder):
                             'title': '400 Bad Request'}
 
 
-def test_multisig_grant_admit(seeder, helpers):
+def test_multisig(seeder, helpers):
```
iFergal, line 366: This test is now even longer, so not sure. But the setup of multi-sigs and credentials being issued is expensive so decided to keep it all to one test if that's OK.

iFergal:
Note - multi-sig admit should retrieve the recipient of admit from the `rp` field (to be renamed to `ri`). Right now it takes the issuer of the ACDC but that only works for issuances and not presentations.

Will update that after Signify-TS has been updated though to always pass the recipient.

2byrds (approved):
LGTM. Adds exchange endpoints for IPEX multisig offer and agree and provides tests.

PR #286, opened 2024-08-28
fix: send multi-sig admit to to discloser not issuer

### LINKED ISSUE #279: Multi-sig IPEX admit should use i field of exn over ACDC issuer
(I can do this work, just making an issue to track it)

When `/ipex/admit` messages are embedded in a `/multisig/exn` messages, the multisig members will send the admit to the issuer of the ACDC (embedded in grant linked to from admit).

It should use the `i` field of the grant message because it might be from a presentation in which case the admit must go to the presenter (discloser) and not the issuer.

iFergal:
Not always the same!

Closes #279

### DIFF
--- a/src/keria/app/ipexing.py
+++ b/src/keria/app/ipexing.py
@@ -145,14 +145,10 @@ class IpexAdmitCollectionEnd:
 
         agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=rec, topic='credential'))
 
-        embeds = grant.ked['e']
-        acdc = embeds["acdc"]
-        issr = acdc['i']
-
         serder = serdering.SerderKERI(sad=admitked)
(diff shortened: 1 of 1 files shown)


2byrds (approved):
LGTM

iFergal:
@lenkan It didn't affect the credential issuance flow. If you were doing a multi-sig admit in response to a grant from an issuer, the credential was always stored because the granter and issuer are the same.

The problem is if it's a multi-sig presentation - the admit was being sent to the issuer instead of the presenter.

But yeah, agreed I'll try to capture it in the tests and open a PR.

kentbull:
So `grant.ked['i']` refers to the discloser of an IPEX Grant and `grant.ked["e"]["acdc"]["i"]` refers. to the issuer of the ACDC?

iFergal:
@kentbull yep!

PR #288, opened 2024-09-01
test: coverage for ipex exn recipient endpoint for multi-sig IPEX

iFergal:
Coverage for #286 

@lenkan Since these are unit tests, they don't rely on real exn messages being passed back and forth over HTTP. The messages are just cloned on the other agent, which is why the tests didn't catch it.

Added some assertions for where the messages would be sent on a real deployment.

### DIFF
--- a/src/keria/app/ipexing.py
+++ b/src/keria/app/ipexing.py
@@ -148,7 +148,7 @@ class IpexAdmitCollectionEnd:
         serder = serdering.SerderKERI(sad=admitked)
         ims = bytearray(serder.raw) + pathed['exn']
         agent.hby.psr.parseOne(ims=ims)
-        agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=[grant.ked['i']], topic="credential"))
+        agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=[admitked['a']['i']], topic="credential"))
         agent.admits.append(dict(said=admitked['d'], pre=hab.pre))
 
         return agent.monitor.submit(serder.pre, longrunning.OpTypes.exchange, metadata=dict(said=serder.said))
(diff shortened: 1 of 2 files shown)


src/keria/app/ipexing.py
```diff
@@ -148,7 +148,7 @@ def sendMultisigExn(agent, hab, ked, sigs, atc, rec):
         serder = serdering.SerderKERI(sad=admitked)
         ims = bytearray(serder.raw) + pathed['exn']
         agent.hby.psr.parseOne(ims=ims)
-        agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=[grant.ked['i']], topic="credential"))
+        agent.exchanges.append(dict(said=serder.said, pre=hab.pre, rec=[admitked['a']['i']], topic="credential"))
```
iFergal, line 151: @2byrds I actually changed this to be more consistent with the rest of the messages. `rec` here determines which service endpoint it will go to, not the recipient (rp) fields of the exn message which is set on the client side.

The effect of relying on `grant.ked['i']` is a client could incorrectly admit it to the wrong prefix, and it would send to the discloser of the grant but the `rp` field wouldn't match. Better to do as the client asked for.

Perhaps we should have some assertions that the admit matches the grant prefix? What do you think?

iFergal: Also note still using `a.i` here instead of `rp` just for consistency. Yet to migrate everything to `rp`

iFergal:
By the way @lenkan there's 2 ways to do multi-sig IPEX.

e.g. for admit
- Create IPEX admit and call `submitAdmit`
- Create `/multisig/exn` with admit embedded and send exn to group members individually with `sendFromEvents`

OR
- Create IPEX admit but don't send
- Create `/multisig/exn` with admit embedded and call `submitAdmit` with the `/multisig/exn`

This PR tackles the second, more efficient route but the Signify-TS tests all seem to use the first method. Not sure what you are using on your end. Might be why no one has spotted this bug until now

lenkan:
> By the way @lenkan there's 2 ways to do multi-sig IPEX.

Thanks for the hint. I have been using the first approach (as that is what the Signify-TS examples were using).

src/keria/app/ipexing.py
2byrds: Sounds good to me @iFergal

2byrds (approved):
LGTM

PR #295, opened 2024-09-21
docs: update readme

sairanjit:
### Summary

- Updated readme Python Version

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # KERIA
+
 [![GitHub Actions](https://github.com/webOfTrust/keria/actions/workflows/python-app-ci.yml/badge.svg)](https://github.com/WebOfTrust/keria/actions)
 [![codecov](https://codecov.io/gh/WebOfTrust/keria/branch/main/graph/badge.svg?token=FR5CB2TPYG)](https://codecov.io/gh/WebOfTrust/keria)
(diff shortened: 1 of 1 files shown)


2byrds:
Hi @sairanjit thank you for reading through the README and suggesting changes. For the purpose of updating the python version could you submit an issue noting the python version issue you found. In this case I think you are pointing out that the version doesn't correlate with what is in the setup.py. We could probably generically point developers there instead of stating it explicitly in the README. I think we can identify a way to note this in a resilient way.

sairanjit:
@2byrds This PR was just to update the Python version that is given in the readme which is `3.12.14` which is not present in the [Python Versions list](https://www.python.org/doc/versions/). So, I changed that to `3.12.4`.

2byrds:
> @2byrds This PR was just to update the Python version that is given in the readme which is `3.12.14` which is not present in the [Python Versions list](https://www.python.org/doc/versions/). So, I changed that to `3.12.4`.

Okay thank you for pointing that out @sairanjit . The PR diff showed changes to 37 different lines. And according to [this search](https://github.com/search?q=repo%3AWebOfTrust%2Fkeria%203.12&type=code) I think the version should be set to `3.12.2`. Could you submit a PR that changes only the python version line in the README (and if you'd like to update the other version references to be consistent throughout the repo).

PR #296, opened 2024-09-23
docs: readme updates

sairanjit:
### Summary

- Updated readme with the correct Python version that is in use to `3.12.2+`


sairanjit:
@2byrds I have updated the python version to `3.12.1+`

PR #300, opened 2024-10-04
upgraded to keri 1.2.0-rc3

2byrds:


setup.py
```diff
@@ -34,7 +34,7 @@
 
 setup(
     name='keria',
-    version='0.2.0-dev3',  # also change in src/keria/__init__.py
```
iFergal, line 37: Just a note that you also need to change it in the Makefile

2byrds: done!

kentbull:
Closing as this is obsoleted by #345 and #348.

PR #302, opened 2024-10-09
feat: return 409 if agent already booted

iFergal:
I think 409 is more appropriate here, and gives it some separation from the various other 400 responses from this endpoint without relying on text responses which are more subject to change.


2byrds (approved):
Seems reasonable "The HTTP 409 Conflict [client error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses) status code indicates a request conflict with the current state of the target resource."

PR #303, opened 2024-10-09
build: docker multi-arch builds for x86 and arm

iFergal:
This is a direct copy of @m00sey's keripy work in WebOfTrust/keripy#832 but for KERIA, hopefully no issues.

Adds arm docker builds and makes them available on the same tag for both architectures. (and resolves the re-publishing of old builds from `--all-tags` push)


2byrds (approved):
'containerd' is funny. Since this is a direct copy from keripy, let's try it out

PR #304, opened 2024-10-10
feat: endpoint to get credential registry state

iFergal:
This allows you to get the vc state of any known credential processed in a Tever. It does not need to be a registry we control.

The state can also be gotten from the credential endpoint if it's issued to us and admitted.

However the motivation for this is to be able to check the status of a credential prior to accepting a credential with IPEX admit. It's mostly just to cover an edge case in our UI, especially if you receive the original IPEX grant for issuance and another when revoking.

It will also support checking the status of untargetted ACDCs that have no issuee.

### DIFF
--- a/src/keria/app/credentialing.py
+++ b/src/keria/app/credentialing.py
@@ -34,7 +34,10 @@ def loadEnds(app, identifierResource):
 
     credentialCollectionEnd = CredentialCollectionEnd(identifierResource)
     app.add_route("/identifiers/{name}/credentials", credentialCollectionEnd)
-    
+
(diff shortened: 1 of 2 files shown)


2byrds (approved):
LGTM. Thank you for the explanation in the PR and for including testing of the endpoint.

iFergal:
thanks @2byrds if we're ok to merge I can create a corresponding PR for Signify! If it conflicts with release cycles I can wait though, no stress.

(I still haven't done the Signify PR for the `/credentials/verify` endpoint but to integration/e2e test that I need my keripy PR to merge... on a fork atm :P)

iFergal:
makes sense to me, thanks @2byrds!

PR #311, opened 2024-10-18
fix: docker image reference cannot contain uppercase characters

### LINKED ISSUE #309: Fix docker tag repo pointer in github action
https://github.com/WebOfTrust/keria/actions/runs/11315384968/job/31466498211

lenkan:
Using `WebOfTrust/keria` as the image name causes docker to choke and try to publish to a docker registry hosted at `WebOfTrust` instead of `docker.io`. This is what happened here: https://github.com/WebOfTrust/keria/actions/runs/11315384968/job/31466498211

Closes #309 

Read more here: https://docs.docker.com/reference/cli/docker/image/tag/

> PATH: The path consists of slash-separated components. Each component may contain **lowercase letters**, digits and separators. A separator is defined as a period, one or two underscores, or one or more hyphens. A component may not start or end with a separator. While the [OCI Distribution Specification](https://github.com/opencontainers/distribution-spec) supports more than two slash-separated components, most registries only support two slash-separated components. For Docker's public registry, the path format is as follows:

### DIFF
--- a/.github/workflows/publish-keria.yml
+++ b/.github/workflows/publish-keria.yml
@@ -1,4 +1,3 @@
-
 name: Publish Docker image
 
 on:
@@ -24,7 +23,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v4
         with:
-          images: WebOfTrust/keria
+          images: weboftrust/keria
 
       - name: Set up Docker buildx
         id: buildx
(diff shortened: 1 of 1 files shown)


rodolfomiranda (approved):

pfeairheller (approved):

iFergal:
@2byrds can we try a dev release again now that this is merged?

2byrds:
> @2byrds can we try a dev release again now that this is merged?

@iFergal what do you want the version to be?

PR #313, opened 2024-10-28
fix: allow cors with falcon 4

iFergal:
Falcon 4 has been released so setup.py is picking this up, and it has this slight interface change.


2byrds (approved):
LGTM

Issue #301, opened 2024-10-08
Available witness OOBI list from KERIA to smoothen UX


iFergal:
One of the user/UX challenges I see with wallets in particular is selection of witnesses and watchers. If every end user needs to know how to select witnesses then it will be a barrier to adoptability.

So I was considering the idea of KERIA announcing known witnesses it controls in its infrastructure and being able to update that list. Usage of non promiscuous witnesses for identifiers could be approved by the KERIA instance. This would lend itself to paid providers for KERIA with low user setup.

This does present some risks, particularly from a compromised KERIA updating the list of known witnesses. Maybe there are some mitigations here such as witnesses needing to sign this "updated list" etc, and having witnesses from other providers.

Curious to hear other people's thoughts.

2byrds:
> One of the user/UX challenges I see with wallets in particular is selection of witnesses and watchers. If every end user needs to know how to select witnesses then it will be a barrier to adoptability.
> 
> So I was considering the idea of KERIA announcing known witnesses it controls in its infrastructure and being able to update that list. Usage of non promiscuous witnesses for identifiers could be approved by the KERIA instance. This would lend itself to paid providers for KERIA with low user setup.
> 
> This does present some risks, particularly from a compromised KERIA updating the list of known witnesses. Maybe there are some mitigations here such as witnesses needing to sign this "updated list" etc, and having witnesses from other providers.
> 
> Curious to hear other people's thoughts.

I like the idea. When you say 'known witnesses it controls' do you simply mean that the company that hosts the KERIA service, also hosts the witness services?
I think it is good that you called out witnesses from other providers since it is generally important but especially worthwhile for greater security, even more so in this case

iFergal:
Yeah, in the most basic case I meant that the company who hosts the KERIA also hosts the witnesses - this is probably not too different to what an enterprise experimenting with integrating KERI would do now in a pilot.

But yes, it would be a lot better to select witnesses from a mix of other providers to de-risk. Especially if witness rotation requires signatures from a threshold amount of prior witnesses.

Maybe this would be like a network of interconnected providers that use each other's witnesses (with some restrictions on load, or appropriate billing). Otherwise an end user would need to select a KERIA provider, and multiple witness providers (..and watchers) and become overwhelming to manage and just decide to revert to something less secure.

SmithSamuelM:
[...]

PR #318, opened 2024-11-08
feat: agent config retrieval

iFergal:
Simple endpoint to retrieve the agent JSON config.

The motivation for this is a really low effort way to discover the witnesses etc that KERIA is aware of when deployed, as described in #301. There might be better solutions that this but it will be OK for my needs for now.

This allows a Signify client to discover from `iurls` witnesses, watchers etc, provided the OOBI URLs are given an appropriate tag as a query param.

I only exposed `iurls` but named it as a config endpoint so more can be exposed in the future if it makes sense.


2byrds (approved):
Nice! LGTM.

PR #330, opened 2024-12-06
feat: delete/remove credential endpoint

### LINKED ISSUE #79: Method to delete credentials from db
We may need a method for receivers of credential to delete the ones that are not longer needed/wanted.

iFergal:
Resolves #79. `DELETE /credentials/{said}` - it will remove the credential from the DB and any associated indices.

This is different to revoke, which is under `DELETE /identifiers/{name}/credentials/{said}` (the path used for issuance).

### DIFF
--- a/tests/app/test_credentialing.py
+++ b/tests/app/test_credentialing.py
@@ -487,6 +487,22 @@ def test_credentialing_ends(helpers, seeder):
         assert res.json == {'description': f"registry EBVaw6pCqfMIiZGkA6qevzRUGsxTRuZXxl6YG1neeCGF not found",
                             'title': '404 Not Found'}
 
+        res = client.simulate_delete(f"/credentials/doesnotexist")
+        assert res.status_code == 404
+        assert res.json == {'description': f"credential for said doesnotexist not found.",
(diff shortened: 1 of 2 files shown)


2byrds (approved):
LGTM

tests/app/test_credentialing.py
```diff
@@ -487,6 +487,22 @@ def test_credentialing_ends(helpers, seeder):
...
 
+        res = client.simulate_delete(f"/credentials/doesnotexist")
+        assert res.status_code == 404
+        assert res.json == {'description': f"credential for said doesnotexist not found.",
+                            'title': '404 Not Found'}
+
+        res = client.simulate_delete(f"/credentials/{saids[0]}")
+        assert res.status_code == 204
+
+        # Check db directly to make sure all indices are gone too (GET endpoints don't cover all indices)
```
lenkan, line 498: @iFergal Perhaps an assertion that you get HTTP 404 (and not for example, a 500), after the deletion could be added here? Just to ensure that the databases are not left in an incomplete state that causes errors.

iFergal: good point, will update

PR #332, opened 2024-12-09
ci(gha): multi-arch docker builds

### LINKED ISSUE #329: Multi-architecture docker builds

I updated the makefile to match keripy but the CICD isn't using the makefile. So either they need to be consolidated or the CICD can pull from the makefile.

iFergal:
To resolve #329. The Makefile is only used for local pushes, so also updating this.

Looking at the output of https://github.com/WebOfTrust/keria/actions/runs/11900767151/job/33162302638 -> `Set up Docker buildx` -> `Inspect builder` -> `"platforms": "linux/amd64,linux/amd64/v2,linux/amd64/v3,linux/386"`, we will need QEMU for arm.

### DIFF
--- a/.github/workflows/publish-keria.yml
+++ b/.github/workflows/publish-keria.yml
@@ -25,6 +25,10 @@ jobs:
         with:
           images: weboftrust/keria
 
+      # For multi-arch
+      - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


2byrds (approved):
LGTM

PR #333, opened 2024-12-09
feat: configure keria with environment variables

### LINKED ISSUE #326: Provide a way to configure keria through environment variables

It would be convenient to be able to configure a keria instance through environment variables instead of having to create configuration files and modifying the entrypoint.

The entrypoint and command in the docker image could be:

```
ENTRYPOINT ["keria"]
CMD ["start"]
```

Then environment variables can control any deviations from the default configuration options. For example:

- `KERIA_PASSCODE` for setting the passcode for the keystore
- `KERIA_CURLS` to configure the `curls` entry from the configuration file, e.g. `https://mykeriainstance.com:3902`.
- `KERIA_ADMIN_PORT` to configure the admin port
- `KERIA_HTTP_PORT` to configure the "http" port
- `KERIA_BOOT_PORT` to configure the "boot" port

The order of precedence for configuration can be:

- Command line option
- Environment variable
- Configuration file

This would allow us to configure a KERIA container without having to create a custom config file and without customizing the entrypoint and command. 

## Current state:

To illustrate the current state, something like the following is currently needed to configure a passcode and agent urls:

```
// config/keria.json
{
    "dt": "2023-12-01T10:05:25.062609+00:00",
    "keria": {
        "dt": "2023-12-01T10:05:25.062609+00:00",
        "curls": ["http://keria:3902/", "http://localhost:3902/"]
    }
}

// config/entrypoint.sh
#!/bin/bash
keria start --config-dir /keria/config --config-file keria --name agent --passcode $PASSCODE --admin $KERIA_ADMIN_PORT

// docker-compose.yaml
services:
    keria:
        image: weboftrust/keria
        environment:
            KERI_AGENT_CORS: 1
            KERIA_ADMIN_PORT=3901
        volumes:
            - ./config/keria.json:/keria/config/keri/cf/keria.json
            - ./config/entrypoint.sh:/keria/entrypoint.sh
        entrypoint: /keria/entrypoint.sh
```

## Suggested use case

To illustrate the proposed use case, I would like the published keria image to support the following:
  
```
// docker-compose.yaml
services:
    keria:
        image: weboftrust/keria
        environment:
            KERI_AGENT_CORS: 1
            KERIA_PASSCODE: <secret injected passcode>
            KERIA_ADMIN_PORT: 3901
            KERIA_AGENT_CURLS: http://localhost:3902/;http://keria:3902/
```

This would make it easier to deploy the published KERIA image to container runtimes (like kubernetes, AWS ECS etc.) without modifications.

## Notes

- I already have some code for this and will open up a PR if people agree. 
- Please share your thoughts about any implications of configuring the "curls" value this way.
- If we do this, what do you think about the `KERI_` vs `KERIA_` prefix of the environment variable names?

lenkan:
- Move all access to os.getenv to `start.py` to make it easier to get an overview of what environment variables are available.
  - Previous ones were KERI_AGENT_CORS and KERIA_RELEASER_TIMEOUT
- Added `KERIA_CURLS` which is the endpoint for the "KERI port". Since this is dependent on the deployment, it is convenient to be able to configure this with an environment variable.
- Added `KERIA_DURLS` and `KERIA_IURLS`, to be able to configure "default" witnesses and data urls for agents.
- Removed the default docker entrypoint that sets up "demo-witness-oobis". From what I could see, this file is not copied into the image anyway.

Closes #326

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,23 @@ All Agent db access is through the associated Agent.
     ```
 
 #### Run with docker
-* Specify an entrypoint with proper configuration, for instance if you want to use the demo-witness-oobis that is under the scripts/keri/cf dir:
+
+* The easiest way to configure a keria container is with environment variables. See below example for a working docker-compose configuration
+
+```yaml
+services:
+    keria:
+        image: weboftrust/keria:latest
+        environment:
+            KERI_AGENT_CORS: 1
+            KERIA_CURLS: http://<keria-hostname>:3902/
+            KERIA_IURLS: http://<witness-demo-hostname>:5642/oobi/BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha;http://<witness-demo-hostname>:5643/oobi/BLskRTInXnMxWaGqcpSyMgo0nYbalW99cGZESrz3zapM
+        ports:
+            - 3901:3901
+            - 3902:3902
+            - 3903:3903
 ```
-ENTRYPOINT ["keria", "start",  "--config-file", "demo-witness-oobis", "--config-dir", "./scripts"]
-```
+
 You can see a [working example here](https://github.com/WebOfTrust/signify-ts/blob/main/docker-compose.yaml).
 
 ### Running Tests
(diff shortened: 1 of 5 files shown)


iFergal (approved):
LGTM!

2byrds (approved):
LGTM

PR #334, opened 2024-12-09
feat: basic protection boot endpoint

### LINKED ISSUE #327: Provide a way to protect the boot endpoint

I suggest there should be a way to protect the boot endpoint with some simple authentication. In my opinion, this is convenient even if the boot port is not exposed publicly on the internet. It would also make it easier for people to spin up a KERIA instance publicly without having to worry about unauthorized people booting agents.

# Suggestion

Implement a [Basic authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication) check by configuring the keria boot endpoint with a `username` and a `password`. Like this:

```
keria start --boot-username username --boot-password password --boot 3903
```

Or, using environment variables

```
export KERIA_BOOT_USERNAME=username
export KERIA_BOOT_PASSWORD=password
keria start --boot 3903
```

To boot an agent, you can then either:

- Use basic auth in the Boot URL: `http://username:password@localhost:3903/boot`, most http clients will set the authorization header for you.
- Set the Authorization header in the boot request manually to `Authorization: Basic dXNlcm5hbWU6cGFzc3dvcmQ=`

We can also add a flag to completely disable the boot endpoint (https://github.com/WebOfTrust/keria/issues/203). E.g.

```
keria start --no-boot-endpoint
```

Or, 

```
export KERIA_DISABLE_BOOT_ENDPOINT=1
```

# Considerations

There are probably much more sophisticated ways to protect the boot endpoint. Please share your thoughts. In my opinion, this could be sufficient and will allow users to more easily deploy keria instances on the internet for testing purposes.

I have code and unit tests for this, so I can open a PR if we can agree this is a good idea.

lenkan:
This PR implements a basic protection of the boot endpoint, this enables you to more easily deploy test instances publicly. This way you can add the authorization to the URL. See rationale in #327

Closes #327

If we also land #333, I will rebase this on top and also expose this feature as command line flags.

I also thought this could be implemented as a falcon middleware. But wanted to get feedback first whether or not it should be included at all.

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -4,6 +4,7 @@ KERIA
 keria.app.agenting module
 
 """
+from base64 import b64decode
 import json
 import os
 import datetime
@@ -53,7 +54,8 @@ logger = ogler.getLogger()
 
 
 def setup(name, bran, adminPort, bootPort, base='', httpPort=None, configFile=None, configDir=None,
-          keypath=None, certpath=None, cafilepath=None, cors=False, releaseTimeout=None, curls=None, iurls=None, durls=None):
+          keypath=None, certpath=None, cafilepath=None, cors=False, releaseTimeout=None, curls=None,
+          iurls=None, durls=None, bootUsername=None, bootPassword=None):
     """ Set up an ahab in Signify mode """
 
     agency = Agency(name=name, base=base, bran=bran, configFile=configFile, configDir=configDir, releaseTimeout=releaseTimeout, curls=curls, iurls=iurls, durls=durls)
@@ -66,7 +68,7 @@ def setup(name, bran, adminPort, bootPort, base='', httpPort=None, configFile=No
     if not bootServer.reopen():
         raise RuntimeError(f"cannot create boot http server on port {bootPort}")
     bootServerDoer = http.ServerDoer(server=bootServer)
-    bootEnd = BootEnd(agency)
+    bootEnd = BootEnd(agency, username=bootUsername, password=bootPassword)
     bootApp.add_route("/boot", bootEnd)
     bootApp.add_route("/health", HealthEnd())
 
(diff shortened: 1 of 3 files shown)


src/keria/app/agenting.py
```diff
@@ -857,17 +858,40 @@ def loadEnds(app):
...
+            password (str): password for boot request
         """
-        self.authn = authing.Authenticater(agency=agency)
+        self.username = username
+        self.password = password
         self.agency = agency
 
-    def on_post(self, req, rep):
+    def authenticate(self, req: falcon.Request):
+        if self.username is None and self.password is None:
```
iFergal, line 873: Maybe we should have `or` instead of `and` in case it's half configured.

src/keria/app/cli/commands/start.py
```diff
@@ -108,6 +109,8 @@ def runAgent(name="ahab", base="", bran="", admin=3901, http=3902, boot=3903, co
                                 configDir=configDir,
                                 keypath=keypath,
                                 certpath=certpath,
-                                cafilepath=cafilepath))
+                                cafilepath=cafilepath,
+                                bootPassword=os.getenv("KERIA_BOOT_PASSWORD"),
```
iFergal, line 113: As we discussed, I would still suggest some indication that a more comprehensive solution should be used in production and that this is more related to test infrastructure. e.g. renaming env vars, some comments on the `authenticate` method etc

lenkan: What about `KERIA_EXPERIMENTAL_BOOT_PASSWORD` for now? That indicates it might change and/or be removed.

iFergal: Works for me!

src/keria/app/agenting.py
lenkan: With the current implementation, you will get a 401 if it is half configured. This check returns without error if _neither_ username or password is configured. If anyone of them is configured, it must match the basic auth values. But perhaps there is a way to make that more clear?

lenkan: Which is arguably OK, because that is how you configured it.

iFergal: Yeah, that's fair enough. Nevermind then

2byrds (approved):
lgtm

PR #335, opened 2024-12-10
feat: expose inception dts in hab info

iFergal:
Exposing the identifier creation time in hab info. Useful for displaying in a UI.

For so far as I can tell this value is always set even when the identifier is still in escrow.

### DIFF
--- a/tests/app/test_aiding.py
+++ b/tests/app/test_aiding.py
@@ -10,6 +10,7 @@ from dataclasses import asdict
 import json
 import os
 import pytest
+from datetime import datetime
 
 import falcon
 from falcon import testing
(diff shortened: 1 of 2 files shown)


tests/app/test_aiding.py
```diff
@@ -1624,6 +1625,9 @@ def test_rotation(helpers):
         assert aid["name"] == "aid1"
         assert aid["prefix"] == "EHgwVwQT15OJvilVvW57HE4w0-GPs_Stj2OFoAHZSysY"
 
+        icp_dt = aid["icp_dt"]
+        datetime.fromisoformat(icp_dt)  # will raise an error if the format is not a valid iso dt
```
iFergal, line 1629: changes every test run so can't check against a literal

2byrds (approved):
lgtm

PR #337, opened 2024-12-11
feat: handle sigterm to allow for graceful shutdown

### LINKED ISSUE #175: How to do graceful shutdown of KERIA
Currently, it seems like KERIA does not handle SIGTERM signals. Most container runtimes would send a SIGTERM signal, followed by a timeout of X seconds/minutes and then send a SIGKILL, which will not allow KERIA to gracefully shutdown.

Are there another recommended way to shutdown a KERIA instance? If not, I suggest to implement a SIGTERM listener that ensures we stop receiving new HTTP requests, then shutting down all the doers, allowing for the process to stop gracefully.

More urgently, do you see any potential issues with sending SIGKILL and killing the process immediately?

Any other thoughts on this?

One way to try this out is to use the docker-compose environment in `signify-ts` repo.

```
docker compose up keria
docker compose kill -s SIGTERM
```

Notice that keria does not shut down on its own. Then do:

```
docker compose kill -s SIGKILL
```

Keria then shuts down immediately.

lenkan:
See #175 for rationale.

This PR implements handling of the SIGTERM, which is sent by runtimes such as docker to the process to allow it to shut down gracefully. I do not know `hio`, but the `Doist` class exposes an `exit` function that seems to do the trick.

Before this change, KERIA would exit with a non-zero exit when receiving a `SIGTERM` signal.

Closes #175

### DIFF
--- a/src/keria/app/cli/commands/start.py
+++ b/src/keria/app/cli/commands/start.py
@@ -8,10 +8,11 @@ Witness command line interface
 import argparse
 import logging
 import os
+import signal
 
+from hio.base import doing
 from keri import __version__
 from keri import help
-from keri.app import directing
 
 from keria.app import agenting
 
@@ -89,8 +90,8 @@ def launch(args):
 
     logger = help.ogler.getLogger()
 
-    logger.info("******* Starting Agent for %s listening: admin/%s, http/%s "
-                ".******", args.name, args.admin, args.http)
+    logger.info("Starting Agent for %s listening: admin/%s, http/%s, boot/%s", args.name, args.admin, args.http, args.boot)
+    logger.info("PID: %s", os.getpid())
 
     agency = agenting.setup(name=args.name or "ahab",
                             base=args.base or "",
@@ -111,8 +112,15 @@ def launch(args):
                             bootPassword=args.bootPassword,
                             bootUsername=args.bootUsername)
 
-    directing.runController(doers=agency, expire=0.0)
+    tock = 0.03125
+    doist = doing.Doist(limit=0.0, tock=tock, real=True)
 
+    def handleSignal(sig, frame):
+        logger.info("Received signal %s", signal.strsignal(sig))
+        doist.exit()
 
-    logger.info("******* Ended Agent for %s listening: admin/%s, http/%s"
-                ".******", args.name, args.admin, args.http)
+    signal.signal(signal.SIGTERM, handleSignal)
+
+    doist.do(doers=agency)
+
+    logger.info("Agent %s gracefully stopped", args.name)
--- a/src/keria/app/cli/commands/start.py
+++ b/src/keria/app/cli/commands/start.py
@@ -1,16 +1,13 @@
 # -*- encoding: utf-8 -*-
 """
 KERIA
-keria.cli.keria.commands module
+keria.cli.keria.commands.start module
 
-Witness command line interface
+KERIA Agent server start command line interface (CLI) command
 """
 import argparse
-import logging
 import os
-import signal
 
-from hio.base import doing
 from keri import __version__
 from keri import help
 
@@ -78,49 +75,34 @@ parser.add_argument("--experimental-boot-username",
                     dest="bootUsername",
                     default=os.getenv("KERIA_EXPERIMENTAL_BOOT_USERNAME"))
 
+
+logger = help.ogler.getLogger()
+
+def launch(args):
+    agenting.runAgency(agenting.KERIAServerConfig(
+        name=args.name or "ahab",
+        base=args.base or "",
+        bran=args.bran,
+        adminPort=args.admin,
+        httpPort=args.http,
+        bootPort=args.boot,
+        configFile=args.configFile,
+        configDir=args.configDir,
+        keyPath=args.keypath,
+        certPath=args.certpath,
+        caFilePath=args.cafilepath,
+        logLevel=args.loglevel,
+        logFile=args.logfile,
+        cors=os.getenv("KERI_AGENT_CORS", "false").lower() in ("true", "1"),
+        releaseTimeout=int(os.getenv("KERIA_RELEASER_TIMEOUT", "86400")),
+        curls=getListVariable("KERIA_CURLS"),
+        iurls=getListVariable("KERIA_IURLS"),
+        durls=getListVariable("KERIA_DURLS"),
+        bootPassword=args.bootPassword,
+        bootUsername=args.bootUsername
+    ))
+    logger.info("Agent %s gracefully stopped", args.name)
+
 def getListVariable(name):
     value = os.getenv(name)
     return value.split(";") if value else None
-
-def launch(args):
-    help.ogler.level = logging.getLevelName(args.loglevel)
-    if(args.logfile != None):
-        help.ogler.headDirPath = args.logfile
-        help.ogler.reopen(name=args.name, temp=False, clear=True)
-
-    logger = help.ogler.getLogger()
-
-    logger.info("Starting Agent for %s listening: admin/%s, http/%s, boot/%s", args.name, args.admin, args.http, args.boot)
-    logger.info("PID: %s", os.getpid())
-
-    agency = agenting.setup(name=args.name or "ahab",
-                            base=args.base or "",
-                            bran=args.bran,
-                            adminPort=args.admin,
-                            httpPort=args.http,
-                            bootPort=args.boot,
-                            configFile=args.configFile,
-                            configDir=args.configDir,
-                            keypath=args.keypath,
-                            certpath=args.certpath,
-                            cafilepath=args.cafilepath,
-                            cors=os.getenv("KERI_AGENT_CORS", "false").lower() in ("true", "1"),
-                            releaseTimeout=int(os.getenv("KERIA_RELEASER_TIMEOUT", "86400")),
-                            curls=getListVariable("KERIA_CURLS"),
-                            iurls=getListVariable("KERIA_IURLS"),
-                            durls=getListVariable("KERIA_DURLS"),
-                            bootPassword=args.bootPassword,
-                            bootUsername=args.bootUsername)
-
-    tock = 0.03125
-    doist = doing.Doist(limit=0.0, tock=tock, real=True)
-
-    def handleSignal(sig, frame):
-        logger.info("Received signal %s", signal.strsignal(sig))
-        doist.exit()
-
-    signal.signal(signal.SIGTERM, handleSignal)
-
-    doist.do(doers=agency)
-
-    logger.info("Agent %s gracefully stopped", args.name)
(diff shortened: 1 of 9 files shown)


src/keria/app/cli/keria.py
```diff
@@ -22,8 +17,7 @@ def main():
         return
 
     try:
-        doers = args.handler(args)
-        directing.runController(doers=doers, expire=0.0)
```
lenkan, line 26: I removed this call because the handlers themselves to not return any `doers`, they are running them themselves.

src/keria/app/cli/commands/start.py
```diff
@@ -111,8 +112,15 @@ def launch(args):
                             bootPassword=args.bootPassword,
                             bootUsername=args.bootUsername)
 
-    directing.runController(doers=agency, expire=0.0)
+    tock = 0.03125
+    doist = doing.Doist(limit=0.0, tock=tock, real=True)
```
lenkan, line 116: The `directing.runController` did not return the `doist`, so there was no way to access it to be able to call `exit`. However, it was only two lines of code, so I thought we could just inline it here.

src/keria/app/cli/keria.py
2byrds: agree https://github.com/WebOfTrust/keria/blob/4981d84a0e6d4536d2a088c1ffdd47871c253799/src/keria/app/cli/commands/start.py#L114

2byrds (approved):
lgtm

lenkan:
Converting to draft because it does not shutdown the agents, only the agency. So it still needs some adjustments. Will pick up again soon.

src/keria/app/cli/commands/start.py
kentbull: This makes sense. This is good enough for now. We could consider making a PR to change KERIpy to have `runController` either return the `Doist` or have a parameter to pass a shutdown signal.

src/keria/app/cli/keria.py
kentbull: I concur, this makes sense.

kentbull (approved):
LGTM

kentbull:
I just tried this with the [vLEI#85](https://github.com/WebOfTrust/vLEI/pull/85) and it worked well.

Yes, we need to send the shutdown signal to the agents, though that should be as simple as making a way to call `.exit()` on their Doer or DoDoer instances.

kentbull:
I added an initial implementation [here](https://github.com/lenkan/keria/pull/1) that calls `Agency.shut` on each agent prior to shutting down the main Doist loop. I added a `doer.close()` to instruct each of the Doers and DoDoers to end their generator loop. If we were to add anything to the shutdown logic to round this out it would be some sort of check on the list of doers that their "done" attribute is "True" which would mean looping in the `Agency.shut` function while checking that the doers have completed.

kentbull:
Looks like the manual `doer.close()` is not necessary because the Doist does this in the `Doist.exit()` function in reverse-nested order, so I removed that code. For reference, here is the code from the Doist class
```python
class Doist(tyming.Tymist):
    def exit(self, deeds=None):
        ...
        if deeds is None:
            deeds = self.deeds

        while(deeds):  # .close each remaining dog in deeds in reverse order
            dog, retime, doer = deeds.pop()  # pop it off in reverse (right side)
            if not dog:  # marker deed
                continue  # skip marker
            try:
                tock = dog.close()  # force GeneratorExit. Maybe log exit tock tyme                <---- THIS IS THE CLOSE
            except StopIteration:
                pass  # Hmm? Not supposed to happen!
            else:  # set done state
                try:
                    doer.done = False  # forced close
                except AttributeError:  # when using bound method for generator function
                    doer.__func__.done = False  # forced close
```

kentbull:
And, really, since this is fully cooperative multitasking I believe calling `Doist.exit()` is sufficient to close all of the agents as well. So calling `agent.close()` per agent should not be necessary because the `Doist` scheduler should do all of that on its own. I will confirm this with Sam and if true will remove that code from my PR to Daniel L's PR.

SmithSamuelM:
For and Doer any resources that need to be opened should be opened inside the Doer's "enter" method and then closed inside the Doer's "exit" method. The Doist when shutting down i.e exiting will call "exit" on each of its Doers and DoDoers and DoDoers will call exit on each of their Doer and DoDoers all that way down.  

The Doist main loop has a try finally that calls its exit method.   It breaks the main loop thereby calling exit on a SigInt (KeyboardInterrupt or Cntl-C, a SystemExit or any other exception.  Some system exits like SigTerm will terminate the process prematurely before the exit method has time to complete. But SigInt will not and should exit cleanly. Therefore  In your dev ops process manager you should terminate Hio using SigInt not SigTerm.  Then you are reasonably assured that all your resources that are closed in Doer. exit() methods will be closed before the process running the Doist terminates.

kentbull:
@SmithSamuelM thank you for clarifying the default behavior of the Doist. That makes sense to me now. We need to move some of the startup logic for both the Agency and the individual Agent instances to the enter and close lifecycle functions in order to leverage the lifecycle context management of HIO. Then graceful shutdown will be as simple as calling Doist.exit().

I will go make that change.

I do support the addition of the SIGTERM handler since the default behavior of both Docker and Kubernetes is to send a SIGTERM, pause for a period (10s and 30s, respectively), and then send a SIGKILL is the process is still alive. SystemD and other process managers that manage daemon-like processes also use SIGTERM so it seems an appropriate choice to trigger the doist.exit() call like @lenkan has written. 

@SmithSamuelM would it make sense to add a SIGTERM signal handler directly to the HIO library or would you prefer the HIO library only handle SIGINT and leave it up to application libraries to decide if they want to handle SIGTERM?

I imagine the broad adoption of SIGTERM by DevOps tools like Docker and K8s may make handling SIGTERM in HIO attractive. Let me know and I'll open a PR there if you would like.

lenkan:
> Some system exits like SigTerm will terminate the process prematurely before the exit method has time to complete.

Are you not mixing up SIGTERM with, for example SIGKILL? I would say that the SIGTERM signal is specifically meant to ask a process to terminate gracefully. Upon receiving it, the process will decide itself when it is time to exit. https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html

I vote for implementing graceful shutdown of keria using SIGTERM. I believe this is the expected behaviour to the vast majority of users. I am not sure I would expect hio to do it for me though. I would rather explicitly tell hio to stop at the keria level.

SmithSamuelM:
Extensive testing history indicates that SigInt works because unlike SigTerm, the python interpreter at a low lever catches SigInt and propogates it through the exception handling hierarchy. While there are corner cases where this may be problematic, the Doist loop is avoiding most of those.  To explicitly handle SegTerm requires using the signals library which is then being run by the interpreter instead of being in the interpreter. This introduces other corner cases. Often the solution to those corner cases looks like, running two processes or threads, one thread 
/process that handles the signals and the other thread/process that runs the code. 

 Phil ran into problems with SigTerm when using supervisord which by default was shutting down with sigterm. When he switched to sigint then the problems went away.  

It may very well be true that the problem of doist.exit() not running to completion is the SigKill not the SigTerm and simply calling registring signals for SigTerm on the Doist main loop may be sufficient without worrying about the corner cases. But it was simply enough for Phil to reconfigure SuperviserD to use SigInt to kill the Doist rather than solve the signals corner cases.

SmithSamuelM:
In the use case where you are terminating KERI is it possible to use SigInt instead of SigTerm?

kentbull:
> signals library which is then being run by the interpreter instead of being in the interpreter.

This is interesting, I did not know this about the signals library.

> It may very well be true that the problem of doist.exit() not running to completion is the SigKill not the SigTerm

In the context of Docker and Kubernetes the Doist main loop did not respond to a SigTerm at all. SigTerm had no effect, which caused the process to remain alive and later be killed with a SigKill.

> registring signals for SigTerm on the Doist main loop may be sufficient without worrying about the corner cases

@SmithSamuelM You mentioned having a Doer in the main loop that checks for the SigTerm on every execution. I added this in [my latest commit](https://github.com/lenkan/keria/pull/1/commits/dc25299fb23baedf8807e088506e6a01a624db17#diff-b3450764caf80c1f90e241cdb445b93cf5a1ce2fe75d7624a5d8c3daa1b70c2dR8) to @lenkan 's fork's branch for this work. Let me know if I did it right or wrong.

I used a `DoDoer` as follows. I tried to use a `Doer` yet when I tried to call `Doist.exit()` from within `Doer.recur()` then it never seemed to have any effect, which means I was using it wrong. I imagine this DoDoer could be simplified to a Doer, I just don't know how to do that.

This DoDoer does work as intended though.
```python
class GracefulShutdownDoer(doing.DoDoer):
    def __init__(self, doist, agency, **kwa):
        self.doist: Doist = doist
        self.agency = agency
        self.shutdown_flag = False

        # Register signal handler
        signal.signal(signal.SIGTERM, self.handle_sigterm)
        signal.signal(signal.SIGINT, self.handle_sigterm)
        logger.info("Registered signal handlers for SIGTERM and SIGINT")

        super().__init__(doers=[self.shutdown], **kwa)

    def handle_sigterm(self, signum, frame):
        logger.info(f"Received signal {signum}, initiating graceful shutdown.")
        self.shutdown_flag = True

    def shutdown_agents(self, agents):
        logger.info("Stopping %s agents", len(agents))
        for caid in agents:
            self.agency.shut(self.agency.agents[caid])

    @doing.doize()
    def shutdown(self, tymth, tock=0.0):
        self.wind(tymth)
        while not self.shutdown_flag:
            yield tock

        # Once shutdown_flag is set, exit the Doist loop
        self.shutdown_agents(list(self.agency.agents.keys()))
        logger.info(f"Shutting down main Doist loop")
        self.doist.exit()
```


@lenkan will you accept [my PR](https://github.com/lenkan/keria/pull/1) to this draft PR? I did change a few more things in order to make unit testing easier for both SigTerm and SigInt. The DoDoer I added listens for both signals and then changes one piece of state, `self.shutdown_flag`, that the Doer checks on each loop iteration.

I also cleaned up the arguments to `agenting.setup` which I renamed to `agenting.setupDoers` since it was getting unwieldy. I added a documented `@dataclass` for those arguments and moved all of the agency setup and runtime logic to the `agenting` module.

I believe we can merge this PR once my PR to Daniel's fork is accepted.

kentbull (approved):
LGTM

SmithSamuelM:
Better would be to just 
```python
raise KeyboardInterrupt
```

Or  to call 

```python
sys.exit()
```

rather than calling doist.exit()
The latter is more brittle as it fixes the dependency on a what the Doist does in response to exiting conditions rather than simply triggering the exiting conditions as the former two examples do.

kmk142789 (approved):

PR #344, opened 2025-01-06
fix: 400 for invalid end role signatures

iFergal:
Feedback to client instead of passing the exception.

Based on the test, passing bad signatures causes `kering.ValidationError` to be thrown from `validateSigs` before `kering.UnverifiedReplyError` is hit.

### DIFF
--- a/src/keria/app/aiding.py
+++ b/src/keria/app/aiding.py
@@ -1330,8 +1330,8 @@ class EndRoleCollectionEnd:
         )
         try:
             agent.hby.rvy.processReply(rserder, tsgs=[tsg])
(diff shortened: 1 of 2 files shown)


2byrds (approved):
LGTM

iFergal:
@2byrds OK, running multi-sig tests this was a mistake. `UnverifiedReplyError` is thrown before all signatures are gathered. Will create a new PR separating the two and bring back the `pass`.

kentbull:
Agreed, yes, I am hitting the same exact problem. `UnverifiedReplyError` is thrown while the end role is in escrow waiting for all signatures to arrive.

PR #346, opened 2025-01-07
fix: pass UnverifiedReplyError again for end role

iFergal:
#344 was wrong because `UnverifiedReplyError` is thrown for insufficient signatures (e.g. multi-sig). Adjusted to only return 400 for invalid signatures.


kentbull (approved):
LGTM

PR #347, opened 2025-01-10
Revert "fix: 400 for invalid end role signatures"

kentbull:
Reverts WebOfTrust/keria#344, as the `UnverifiedReplyError` is a part of the normal operations of the reply message escrows and should not result in a 400 error being sent to the SignifyTS client.


iFergal:
@kentbull please see the PR i opened last week to correct: #346 

This reverts `UnverifiedReplyError` but keeps the new logic to make sure the signature is correct.

kentbull:
oh, okay, I'll go take a look at that and close this one.

PR #345, opened 2025-01-07
build: 0.2.0 release candidate (latest 1.2.2 keripy)

iFergal:
Latest released version, in preparation for a `0.2.0` release of KERIA.

PR #348, opened 2025-01-10
chore: upgrade deps to prep for 0.2.0-rc1 release

kentbull:
This does a few things:
1. bumps version to 0.3.0 for the next release.
2. Upgrades to KERI 1.2.2 (only had to change the `getFelItemPreIter` return tuple)
3. Adds a smaple KERIA config at `scripts/keri/cf/keria.json` with sample tock configuration.
4. Updates all dependencies to the latest versions in PyPi.
5. Uses specific versions in setup.py so builds are strictly repeatable.
6. adds a little more project metadata

### DIFF
--- a/scripts/keri/cf/keria.json
+++ b/scripts/keri/cf/keria.json
@@ -1,0 +1,16 @@
+{
+  "dt": "2025-01-03T16:08:30.123456+00:00",
+  "keria": {
+    "dt": "2025-01-03T16:08:30.123457+00:00",
+    "curls": ["http://127.0.0.1:3902/"]
+  },
+  "iurls": [
+    "http://127.0.0.1:5642/oobi/BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha/controller?name=Wan&tag=witness",
+    "http://127.0.0.1:5643/oobi/BLskRTInXnMxWaGqcpSyMgo0nYbalW99cGZESrz3zapM/controller?name=Wil&tag=witness",
+    "http://127.0.0.1:5644/oobi/BIKKuvBwpmDVA4Ds-EpL5bt9OqPzWPja2LigFYZN2YfX/controller?name=Wes&tag=witness"
+  ],
+  "tocks": {
+    "initer": 0.0,
+    "escrower": 1.0
+  }
+}
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ Best practices for setup.py and requirements.txt
 https://caremad.io/posts/2013/07/setup-vs-requirement/
 """
 
-
+from pathlib import Path
 from glob import glob
 from os.path import basename
 from os.path import splitext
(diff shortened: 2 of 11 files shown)


iFergal:
@kentbull OK, this will supersede #345 - thanks for the extra changes.

However, we haven't released `0.2.0` yet - we're still on dev tags. Hence, I think this PR should upgrade to `0.2.0-rc1` and create a release branch. And then a separate PR can upgrade to `0.3.0-dev0`.

scripts/keri/cf/keria.json
iFergal, line 1: See https://github.com/WebOfTrust/keria/commit/912e68144ee0f31536c77d67f732ee82469d210c

@lenkan has changed it to use the env instead of this file.

setup.py
```diff
@@ -34,10 +34,11 @@
 
 setup(
     name='keria',
-    version='0.2.0-dev6',  # also change in src/keria/__init__.py
+    version='0.3.0',  # also change in src/keria/__init__.py
     license='Apache Software License 2.0',
     description='KERIA: KERI Agent in the cloud',
-    long_description="KERIA: KERI Agent in the cloud.",
+    long_description=open("README.md").read(),
```
iFergal, line 40: Is this a standard practice, and is it OK with all the images? (I guess so with `text/markdown`)

setup.py
```diff
@@ -64,34 +65,33 @@
         'Topic :: Utilities',
     ],
     project_urls={
+        'Documentation': 'https://weboftrust.github.io/keridoc',
         'Issue Tracker': 'https://github.com/WebOfTrust/keria/issues',
+        'Source': 'https://github.com/WebOfTrust/keria/issues',
```
iFergal, line 70: This URL is for issue tracker

setup.py
```diff
@@ -34,10 +34,11 @@
 
 setup(
     name='keria',
-    version='0.2.0-dev6',  # also change in src/keria/__init__.py
+    version='0.3.0',  # also change in src/keria/__init__.py
```
iFergal, line 37: Already left a comment on the version itself, but you're missing the change to the `Makefile`

scripts/keri/cf/keria.json
kentbull: The keria.json file is still used though, and the tocks are only configurable with the JSON file right now.
It would be worth talking over which approach we like better. The `keria.json` file was intended to be a bootstrap config file for all of the configuration needs of an agent.

And this file is just to make sure the example keria.json file is complete and correct.

iFergal: OK fair, it's just an example and not used by the Dockerfile. That's OK - I just wanted to avoid having both and it being confusing. Tocks should probably be configurable by the env too then.

Aside from that, the env can bootstrap the configuration now in the same way.

Probably should be documented.

setup.py
kentbull: It is a nice way to get the Readme to show up on the PyPi package page.
Here's the [docs](https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/) on how to do it. I found a better way using [this SO answer](https://stackoverflow.com/a/69655159/2084253).

setup.py
kentbull: whoops, good catch, copy paste error.

kentbull:
@iFergal sounds good to me, i's now changed to `0.2.0-rc1`. I also updated versions in the build pipeline and added the most recent Alpine image.

To slightly optimize image size I swapped out alpine-sdk for gcc. Only the gcc dependency is needed in order to properly link libsodium. If it weren't for linking libsodium then we wouldn't need gcc, which was the entire reason alpine-sdk was included.

setup.py
kentbull: This was changed to 0.2.0-rc1 in the most recent commit (force pushed).

scripts/keri/cf/keria.json
kentbull: I'll add docs for this.

iFergal (approved):
LGTM. Not sure if you're still going to update any docs on the env vs config file but not critical.

2byrds (approved):
LGTM

kentbull:
@iFergal see the docs I added in the PR. You can `cd docs/ && make html && make open` to view the docs. I changed the theme to the ReadTheDocs theme which is nice on the eyes.

iFergal (approved):
LGTM!

PR #349, opened 2025-01-13
fix: basic boot password should be able to contain colon characters

lenkan:
Fixes an issue that did not allow `--experimental-boot-password` to contain `:` characters.


2byrds (approved):
LGTM

Issue #287, opened 2024-08-30
Encryption in the KERIA Signify API (and lack of signatures in places currently)


iFergal:
Right now our usage of RFC-9421 has some problems:
- Query params are not signed - https://github.com/WebOfTrust/signify-ts/issues/215
- The body is also not signed.

Generally the body from Signify to KERIA is something that's already signed like an inception event, but I don't think there's anything signing response bodies from KERIA for the likes of e.g. notification arrays etc.

---

Another general issue is none of this is encrypted, except for if we use TLS. Ideally, we could use something like ESSR here.

Phil has mentioned to me that 1Password uses JWE tokens (so encrypted) in request bodies. We could do something similar except I would argue for full confidentiality and privacy (without relying on TLS) that all the information must be contained in the body. (so one single path for all requests, more like [kapi routing](https://github.com/WebOfTrust/kapi/blob/main/kapi.md))

--

I haven't fully thought about this but would like to kick off a general discussion. At minimum those extra items need to be signed and rely on TLS for encryption but long term I think we should do better than TLS.

dhh1128:
tagging @da-tychinin, who has worked on this and has some opinions; also @rodolfomiranda

rodolfomiranda:
Additionally, on the verification side I think that we need to validate the timestamp to prevent replay attacks.

iFergal:
Need to do some investigation but some cases may already be protected from replay attacks if KERIA does nothing with a request containing the same SAID. Might be wrong.

May be also better in general to work towards something like that, as a timestamp only protects against replay attacks outside of the accepted timestamp window and not within. Good to have both though as a timestamp check is fast so would help with DDoS protection too.

rodolfomiranda:
Ideally, we should check that the timestamp is within a window, but also that is greater than the previous timestamp received for the same call.

iFergal:
Right but would require storing or caching the timestamps of previous calls, where it might be cleaner if our endpoints were designed to just be idempotent based on the SAID of the Signify request.

And the timestamp window could help as a quick sanity check for DoS reasons

iFergal:
**From today's call:** discussed how we could completely migrate to just having ESSR payloads in the body (with the unencrypted ESSR payload in the headers for load balancers). KERIA and Signify would effectively just use 1 endpoint then and do routing (like agent to agent messages) instead of REST. Phil suggested you could even completely remove HTTP and only have a tcp handler for the messages.

[...]

Issue #222, opened 2024-04-04
Provide support for passcode rotation


lenkan:
As a user of a Keria Cloud Wallet, I need to be able to rotate my passcode so that I can change the passcode if there is a risk that my passcode has been compromised.

Currently, there is no documented way to do this as far as I can see. There as been a discussion on the community discord a while back: https://discord.com/channels/1148629222647148624/1148734044545228831/1164288912500461729

There is a function in signify-ts (https://github.com/WebOfTrust/signify-ts/blob/development/src/keri/app/clienting.ts#L360) named rotate that accepts a new passcode. However, it triggers Internal Server Error in KERIA when called. Also a bit unclear what needs to be provided in the second parameter if the intention is to rotate a passcode.

In kli from `keripy`, we have the `kli passcode set` command that supports passcode rotation: https://github.com/WebOfTrust/keripy/blob/0c94c462a44c46f97a6c5d7e683fb9bf821154ba/src/keri/app/cli/commands/passcode/set.py

rodolfomiranda:
the feature was there initially, but probably is broken now. There were a couple of scripts in `Signifypy` that test passcode rotation and passcode rotation 1000 times. The thing with passcode rotation is that you need to regenerate and submit all new encrypted keys, that probably what should go in the second parameter.

2byrds:
From our keri dev meeting:
Code is there but some integration testing and fixes are needed. Signify-ts has the call but KERIA gives a 500.

2byrds:
I have fixed this issue as part of a wave of SIGNIFY-TS testing and fixes related to multisig-vlei-issuance fixes that will include multisig delegation fixes as well https://github.com/WebOfTrust/keria/issues/249.
We are approaching a new Keripy/KERIA release (per the dev discussion today)

2byrds:
Here are the current calls to rotate in siginfy-ts tests https://github.com/search?q=repo%3AWebOfTrust%2Fsignify-ts+rotate+path%3A%2F%5Eexamples%5C%2Fintegration-scripts%5C%2F%2F&type=code
They are dominated by calls with only the aid and empty args (salty rotations).

[...]

PR #351, opened 2025-01-15
feat: ESSR protected client APIs

iFergal:
This PR adds an alternative mechanism to authenticate with the KERIA API from a Signify client, and leaves the current RFC-9421 signed headers in place too. Beyond confidentiality, it also resolves other issues from before such as unsigned query params and bodies. (#287)

The unsigned HTTP request from the client is converted to a HTTP bytestream and embedded in an ESSR payload. This payload becomes the body of a wrapper HTTP request for the `"/"` path - so tunneled via ESSR. The wrapper could be handled by pure TCP but for now this works quite nicely in the Falcon middlewares with minimal changes.

_Because nothing else uses `"/"` as the path, the authentication type is determined based on the path. A cleaner approach in a TCP world would just to have a CESR based command signed and optionally encrypted following ESSR, rather than complicating things with managing headers, or embedding HTTP requests within each other._

I have written the corresponding code in Signify and all integration tests pass.

----------

**Flow for generating request from Signify:**
1. Create HTTP request - e.g. `POST /identifiers <body>`
2. Add `Signify-Resource` header **(Encrypt Sender)**
3. Serialize HTTP request as a HTTP string
4. `crypto_box_seal(http_request_string, public_key_of_keria_agent)`
5. Create a new HTTP request to `POST /` where the body of the request is the sealed box bytes (`application/octet-stream`)
6. Add `Signify-Resource`, `Signify-Timestamp` and `Signify-Receiver` headers (receiver is **Sign Receiver**)
7. Sign the request and add as a header

**Note:**
Like the keripy ESSR parser, the commitment to unsigned HTTP request is done by taking a digest:
```
payload = dict(
    src=<sender>,
    dest=<receiver>,
    d=coring.Diger(ser=cipher, code=MtrDex.Blake3_256).qb64,
    dt=dt,
)
```

A signature is created over this dict. I'm re-using the signature style of the signed headers.

ESSR drawio


iFergal:
https://github.com/WebOfTrust/signify-ts/pull/304 created on Signify side

src/keria/core/authing.py
```diff
@@ -153,59 +149,89 @@ def __init__(self, agency, authn: Authenticater, allowed=None):
...
+        "SERVER_PORT": splitHost[1] if len(splitHost) > 1 else ("433" if splitUrl.scheme == "https" else "80"),
+        "SERVER_PROTOCOL": protocol,
+        "PATH_INFO": splitUrl.path,
+        "QUERY_STRING": splitUrl.query,
+        "CONTENT_TYPE": headers.get("content-type", ""),
+        "CONTENT_LENGTH": str(len(body)) if body else "0",
+    }
+
+    for key, value in headers.items():
+        key = "HTTP_" + key.replace("-", "_").upper()
```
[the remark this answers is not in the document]
iFergal: It's how WSGI (falcon) expects it in the environ that builds the request. The headers in the actual HTTP request still have dashes like normal.

hio also does it: https://github.com/ioflo/hio/blob/main/src/hio/core/http/serving.py#L737-L740

src/keria/core/authing.py
```diff
@@ -234,5 +236,12 @@ def serializeResponse(protocol: str, response: falcon.Response):
         f"{key}: {value}" for key, value in response.headers.items()
         if key.lower() not in CORS_HEADERS
     ])
-    body = response.data.decode("utf-8") if response.data else ""
+
+    if response.text:
+        body = response.text.strip()
```
iFergal, line 241: `text` sometimes used for e.g. 404 responses. None of the signify tests caught this. Need to at least expand the tests in this PR to cover different all the cases for bodies. todo.

src/keria/app/aiding.py
```diff
@@ -245,7 +245,8 @@ def on_put(self, req, rep, caid):
         ctrlHab = agent.hby.habByName(caid, ns="agent")
         ctrlHab.rotate(serder=rot, sigers=[core.Siger(qb64=sig) for sig in sigs])
 
-        if not self.authn.verify(req):
+        # @TODO - foconnor: Not sure if this should be here - Signify is not signing headers for passcode rotation.
+        if not self.authn.inbound(req):
```
iFergal, line 249: @pfeairheller Passcode rotation (`PUT /agent/{caid}`) in Signify doesn't use signed headers (just has a signed body), like the rest of the `/agent` endpoints, so perhaps this shouldn't be checking for signed headers here. If not, I can update Signify to use the correct fetch, and adjust this to accept ESSR if needed.

Any idea @rodolfomiranda?

src/keria/core/authing.py
```diff
@@ -92,120 +156,284 @@ def verify(self, request):
...
+            rep (Response): Falcon response object
 
 
         """
-
         for path in self.allowed:
             if req.path.startswith(path):
                 return
 
-        req.path = quote(req.path)
```
iFergal, line 170: The path is already encoded when sending over the wire. e.g. set `'alice'` to `'alice test'` in a test, and it will fail on main, as `/alice%20test` is being turned into `/alice%2520test` with this line (double encoded).

kentbull:
@iFergal will you rebase this and fix the conflicts? I want to review and get this merged.

iFergal:
@kentbull Conflicts fixed.

kentbull:
Reviewing and will merge this week unless I have changes to request.

src/keria/app/aiding.py
kentbull: SignifyTS should be signing headers for passcode rotation. SignifyPy does. This is just a gap in SignifyTS. Yet I won't hold up merge for this since #222 is where we will track the effort to fix passcode rotation, including for SignifyTS.

src/keria/core/authing.py
kentbull: Agreed, good catch.

kentbull (approved):
I acknowledge `AgentResourceEnd.on_put` passcode rotation is still broken as signed-header requests reaching the `PUT /agent/{caid}` controller-rotation branch fail due to SignifyTS not sending a signed request.

This PR adds a further breakage because `self.authn.inbound(req)` returns None on successful authentication yet the if statement takes that and says `if not None` then error, which is the opposite of what it should be.

The correct code would look like the following, which I will add in a follow-on PR:

```py
try:
    self.authn.inbound(req)
except (kering.AuthNError, ValueError):
    raise falcon.HTTPForbidden(...)
```

I'll approve, merge this, and then set up that follow-on PR.

kentbull:
@iFergal I opened https://github.com/WebOfTrust/keria/pull/451 in response to a regression introduced by the change in meaning of the `authn.inbound()` method compared with the old `authn.verify()`. When verify/inbound returns None that is a clean validation result and the prior if statement incorrectly interpreted None as truthy and raised an error. That is the inverse of what needs to occur. #451 fixes that.

PR #352, opened 2025-01-17
feat: api to add new loc scheme

iFergal:
This allows us to add new location/URL entries for endpoint identifiers that have a role.

I'll be using it to allow an issuer to add a schema discovery endpoint using the new `indexer` role - WebOfTrust/keripy#920 - with itself as the `eid`. So the URL can be retrieved via the base OOBI `http://keria.com/{controllerPre}`. (not `.../agent/{agentPre}`)

This can be extended to have a proper way to retrieve the endpoint given the OOBI, but for my purposes right now it's OK to just parse the OOBI GET response because it's just for fetching SAIDs.


2byrds (approved):
Lgtm

PR #358, opened 2025-03-03
fix: delete keria specific indexes when deleting a credential

iFergal:
When I added the endpoint to delete a credential by SAID from DB (not revoke, just wipe it), I deleted the simple indexes used by keripy. What I missed was the `agenting.Seeker` will create much more sophisticated indexes based on the schema.

I missed this because `/credentials/query` without a filter doesn't search these indexes, only when you pass something more specific like searching based on issuer, attribute etc.

And if it finds it by this indexed search, it can't return it because `cloneCred` will throw an error.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -101,7 +101,7 @@ setup(
         'pytest>=8.3.4',
     ],
     setup_requires=[
-        'setuptools==75.8.0'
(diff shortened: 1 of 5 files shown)


2byrds (approved):
LGTM

PR #363, opened 2025-05-27
feat: add hab name and prefix partial to log messages

kentbull:
This PR adds the following types of log message enhancements:
- uses the `keria:` prefix for log messages instead of `keri:`
- uses the asctime, levelname, and message log fields

For simpler multisig debugging
- some debug logs to ExchangeSender
- Within `IdentifierResourceEnd` adds logs starting with `logger.info("[%s | %s...%s]:` that use the name and AID of the hab, first and last 4 characters of the AID.
- Within `Registrar` adds `logger.info("[%s | %s...%s]:` style logs
- Within `Anchorer` adds `logger.info("[%s | %s...%s]:` style logs
- Within `MultisigRequestCollectionEnd` adds `logger.info("[%s | %s...%s]:` style logs
- Within `ExchangeCollectionEnd` adds `logger.info("[%s | %s...%s]:` style logs


src/keria/app/aiding.py
```diff
@@ -834,6 +834,8 @@ def rotate(agent, name, body):
                 title="invalid rotation",
                 description=f"required field 'rot' missing from request",
             )
+        serder = serdering.SerderKERI(sad=rot)
+        logger.info("[%s | %s]: Rotation event sn=%s SAID=%s", hab.name, hab.pre, serder.sn, serder.said)
```
iFergal, line 838: This prints the full hab pre and not truncated right?

src/keria/app/serving.py
```diff
@@ -50,16 +58,17 @@ def recur(self, tock=0.0):
...
     def exit(self):
         """
         Exits the Doist loop.
         Lifecycle method called once when the Doist running this Doer exits the context for this Doer.
         """
-        logger.info(f"Shutting down main Doist loop")
-        self.doist.exit()
+        logger.info("Shutting down all agents in the Agency")
+        self.shutdown_agents(list(self.agency.agents.keys()))
+        raise KeyboardInterrupt("Graceful shutdown finished, causing Doist loop to exit")
```
iFergal, line 74: For my understanding, why do we need to move shutdown here and raise a keyboard interrupt now?

src/keria/peer/exchanging.py
```diff
@@ -104,6 +106,9 @@ def on_post(req, rep, name):
         serder = serdering.SerderKERI(sad=ked)
         sigers = [core.Siger(qb64=sig) for sig in sigs]
 
+        logger.info("[%s | %s] ExchangeCollectionEnd: route %s received exn %s being sent to [%s]",
```
iFergal, line 109: I think `ExchangeCollectionEnd` or something similar is already in the logs

src/keria/app/agenting.py
```diff
@@ -120,11 +140,11 @@ class KERIAServerConfig:
...
     """Runs a KERIA Agency with the given Doers by calling Doist.do(). Useful for testing."""
-    help.ogler.level = logging.getLevelName(config.logLevel)
-    logger.setLevel(help.ogler.level)
+    ogler.level = logging.getLevelName(config.logLevel)
+    logger.setLevel(ogler.level)
     if config.logFile is not None:
-        help.ogler.headDirPath = config.logFile
-        help.ogler.reopen(name=config.name, temp=False, clear=True)
+        ogler.headDirPath = config.logFile
+        ogler.reopen(name="keria", temp=False, clear=True)
```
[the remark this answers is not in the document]
kentbull: `config.name` should never change as it represents the name of the overall binary, that being `keria`, which functions as a tag in the logs. This identifies a log message as specifically from the `keria` binary. I made a mistake when I set the name to `config.name` because that would make the log message reflect the name of the local keystore (Habery) rather than the tag `keria`.

src/keria/app/agenting.py
```diff
@@ -865,20 +1104,28 @@ def recur(self, tyme=None):
...
         if not self.agentHab.inited:
             return False
-
-        print("  Agent:", self.agentHab.pre, "  Controller:", self.caid)
+        agent_label = f"  Agent Pre: {self.agentHab.pre}, Controller: {self.caid}"
+        # if log level not info or lower just print
+        if logger.level > logging.INFO:
+            print(agent_label)
+        else:
+            logger.info(f"  Agent Pre: {self.agentHab.pre}, Controller: {self.caid}")
```
iFergal, line 1123: If we always want it printed, maybe `print` is OK. Or just set it to only print for info

src/keria/app/agenting.py
```diff
@@ -424,26 +495,156 @@ def lookup(self, pre):
...
+        Called once per agent since self.remove() calls self.exit() when cleaning up each agent.
+        Should only trigger the Doist loop to exit once all agents have been removed.
+        """
+        super(Agency, self).exit(deeds=deeds if deeds else self.deeds)
+        if len(self.agents) == 0:
+            raise KeyboardInterrupt("Agency shutdown complete. Exiting Agency.")
 
 class Agent(doing.DoDoer):
     """
+    An always-on agent for a remote Signify controller of keys at the edge. The top level, DoDoer task object
```
iFergal, line 537: They are only on if the client connected recently, after the last restart

src/keria/app/agenting.py
```diff
@@ -574,6 +783,30 @@ def inceptExtern(self, pre, verfers, digers, **kwargs):
...
+            self.seeker.close(clear=False)
+            self.exnseeker.close(clear=False)
+            self.monitor.opr.close(clear=False)
+            self.notifier.noter.close(clear=False)
+            self.rep.mbx.close(clear=False)
+            self.registrar.rgy.close()
+            self.mgr.rb.close(clear=False)
+            self.hby.close(clear=False)
+        except lmdb.Error as ex:  # Sometimes LMDB will throw an error if the DB is already closed
+            logger.error(f"Error closing databases for agent {self.caid}: {ex}")
```
iFergal, line 806: I only noticed this now, but if there's an error shutting down e.g. `ks`, this means we skip the rest.

It's probably better if we have a list of these to close, and iterate on them with a try catch within the loop. What do you think?

scripts/keria.json
iFergal, line 1: Where is this file used?

src/keria/app/serving.py
iFergal: @kentbull I'm still unsure on why the changes for the shutdown were needed. Is there a specific issue being fixed?

kentbull: This is more of an enhancement than a specific fix that targets modular setup and shutdown logic for agents that provide self-contained context and resource management. In addition to having agents shut themselves down being idiomatic Hio it is also just a good practice. Agents know their own context the best and thus should manage themselves.

The possible problems that could occur when an agent doesn't shut itself down could be race conditions between the thing shutting the agent down and the agent itself and resources possibly not being cleaned up properly. I wouldn't expect HTTP requests to be terribly negatively affected since the Falcon HTTP server is at the Agency level, not the agent level.

Generally it is a good practice to have each component maintain its own knowledge. This conforms to the dependency inversion principle from SOLID where only an agent has to worry about shutting itself down rather than the agency or another doer having to worry about or know how to do that.

kentbull (commented):
Submitting my review.

src/keria/app/aiding.py
kentbull: Ah, I was unclear on how the Github review system worked. I just submitted my review.

src/keria/app/serving.py
kentbull: Also, to the specific question from earlier,
> For my understanding, why do we need to move shutdown here and raise a keyboard interrupt now?
The KeyboardInterrupt must be raised by some part of the shutdown infrastructure so that the Doist running the event loop will shut everything down. Otherwise the Doist will run forever since it is running with Doist.do and there are other tasks, Doers, that run indefinitely. 

Some of the other Doers that run indefinitely include the bootServerDoer, adminServerDoer, httpServerDoer, which should probably shut down before the Agency shuts down. We should probably look at shutting those down first in a future PR.

src/keria/app/agenting.py
kentbull: Adding a timestamp around the message seemed valuable, and I didn't want to make the log message disappear for people who didn't turn INFO log messages off. When I use INFO log levels I also want to avoid duplicate logs.

src/keria/app/agenting.py
kentbull: Good catch. I changed it to "network accessible"

src/keria/app/agenting.py
kentbull: I just added a for loop around the databases in the latest version of the code.

src/keria/app/agenting.py
```diff
@@ -424,26 +495,156 @@ def lookup(self, pre):
...
+    def recur(self, tyme=None, tock=0.0):
+        """
+        Checks once per loop to see if the Agency should shutdown.
+        If so, it will shut down each agent and then exit the Agency by returning True for the task (DoDoer) completion status.
+        """
+        if self.shouldShutdown and len(self.agents) == 0:
+            logger.info("Agency shutdown complete. Exiting Agency.")
+            return True
+        if self.shouldShutdown and len(self.agents) > 0:
+            self.shutdownAgency()
```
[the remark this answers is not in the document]
kentbull: Yes, the reason `True` is returned is to indicate the Agency task is done so that the containing doist will call the .exit() function on this Agency. The .exit() function contains the KeyboardInterrupt which will then signal, like an external Control+C, to the containing Doist that it should close all of its other tasks and shut down.

The KeyboardInterrupt is called from the `exit` instead of the `recur` so that the Agency and any sub-doers all properly call their .exit functions which is the first code line in Agency.exit:
```python
class Agency(doing.DoDoer):
    ...
    def exit(self, rdeeds=None, deeds=None):
        """
        Called once per agent since self.remove() calls self.exit() when cleaning up each agent.
        Should only trigger the Doist loop to exit once all agents have been removed.
        """
        super(Agency, self).exit(deeds=deeds if deeds else self.deeds)  # <-- this line ensures this DoDoer and all child Doers exit by having their .exit called.
        if len(self.agents) == 0:
            raise KeyboardInterrupt("Agency shutdown complete. Exiting Agency.")
```

src/keria/app/serving.py
iFergal: OK, understand the point on agents shutting themselves down. I'd be interested if there's a cleaner way in HIO to exit rather than raising a KeyboardInterrupt - like some hook that Hio could provide or something. But works for me for now

kentbull: If I recall correctly Sam is the one who told me to raise a KeyboardInterrupt as he said that is the idiomatic way in Python, and HIO, to pass a Ctrl-C interrupt signal up to the root scheduler.

src/keria/app/agenting.py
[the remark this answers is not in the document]
kentbull: The reason `True` is not returned from the branch on line 522 is because the process is not done until the count of agents `len(self.agents)` is zero, which is why the if statement says that. So the first branch is only ever triggered once all agents have been shut down. The second branch executes to shut all agents down.

The first branch returning `True` is what causes the `Agency.exit()` function to be called which in turn is caught by the Doist which in turn calls `.exit()` on all of its Doers and shuts down the Doist loop.

iFergal (approved):

setup.py
```diff
@@ -84,13 +84,14 @@
...
         'hio==0.6.14',
-        'keri==1.2.4',
+        'keri==1.2.6',
         'mnemonic==0.21',
         'multicommand==1.0.0',
         'falcon==4.0.2',
         'http_sfv==0.9.9',
         'dataclasses_json==0.6.7',
         'apispec==6.8.1',
+        'deprecation==2.1.0',
```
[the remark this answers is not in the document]
kentbull: This is used to enable the `@deprecated` notice on [keria.app.agenting.Agency.shut](https://github.com/WebOfTrust/keria/pull/363/files#diff-a169faf61f8a8fdf3be4d9d2a164f646003d0f5203efb18f599f4b59393327efR417).

I liked the `@deprecated` construct from the Java world and thought it might be useful here to indicate methods or functions that are superseded by other functions.

src/keria/app/agenting.py
```diff
@@ -865,20 +1105,28 @@ def recur(self, tyme=None):
...
+    """Prints a message once an agent is initialized."""
     def __init__(self, agentHab, caid, tock=0.0):
         self.agentHab = agentHab
         self.caid = caid
         self.tock = tock
         super(Initer, self).__init__(tock=self.tock)
 
-    def recur(self, tyme):
-        """ Prints Agent name and prefix """
+    def print_agent(self):
```
lenkan, line 1115: This does not follow the camelCase naming convention.

src/keria/app/cli/commands/start.py
```diff
@@ -6,8 +6,10 @@
 KERIA Agent server start command line interface (CLI) command
 """
 import argparse
+import logging
```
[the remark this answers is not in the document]
kentbull: I'll remove this on a future PR.

src/keria/app/agenting.py
kentbull: I did not change this in the PR, just the spacing changed.

src/keria/app/agenting.py
```diff
@@ -354,14 +406,21 @@ def create(self, caid, salt=None):
...
     def delete(self, agent):
+        """Deletes the agent from the agency and cleans up its resources."""
         self.adb.agnt.rem(key=agent.caid)
         agent.hby.deleteHab(agent.caid)
         agent.hby.ks.close(clear=True)
         agent.hby.close(clear=True)
 
         del self.agents[agent.caid]
 
+    @deprecated(deprecated_in="0.2.0-rc2", removed_in="1.0.0",
```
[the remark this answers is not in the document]
kentbull: I'll remove that in a PR.

PR #365, opened 2025-05-27
fix: send agent and controller KEL on IPEX grant

kentbull:
This resolves the long-standing double present bug where presenting the first credential from KERIA would fail because the destination did not have the agent KEL.

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -675,8 +675,23 @@ class ExchangeSender(doing.DoDoer):
 
 
 class Granter(doing.DoDoer):
+    """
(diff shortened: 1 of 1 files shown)


iFergal (approved):
Makes sense, would be good to know why it was passing on the second presentation before though

kentbull:
I do not yet know why the second presentation would work. I imagine the Agent KEL is sent somewhere, so this is likely an ordering and escrow processing problem.

iFergal:
In the workflow you are describing, Sally has not resolved the OOBI of the presenter right?

If so, it makes sense that this causes an issue because the Signify and agent identifiers are delegated. Just not sure where the Signify+Agent KELs are ever sent, aside from this PR and the OOBI response.

PR #366, opened 2025-06-04
chore: bump version to 0.2.0 for release to PyPi and Docker HUb

kentbull:
Bumps version to 0.2.0


kentbull:
@lenkan @iFergal it seems like we're ready to do a 0.2.0 release. We have only done release candidates for 0.2.0 and I believe we're ready for a 0.2.0.

Do you have everything you want in the 0.2.0? We should probably add release notes.

PR #385, opened 2025-09-22
chore: add ruff linting

lenkan:
Adds ruff linting with default rules (https://docs.astral.sh/ruff/rules/)

> If you're just getting started with Ruff, the default rule set is a great place to start: it catches a wide variety of common errors (like unused imports) with zero configuration.

- Removes flake8 checks to avoid having conflicting reports in the future.
  - I can add an equivalent ruff check to print warnings like this did before - but I don't see any value in it if it doesn't stop the build
     ```
     # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
     flake8 . --count --exit-zero --ignore=E7,F841,E301,E302,E303 --max-complexity=10 --max-line-length=127 --statistics
     ```
- Uses extras_require.dev to specify dev dependencies - I don't know if this is common way to specify development dependencies, but I found some examples where that is done. Please advice.
- All changes to source code are auto-fixed except: https://github.com/WebOfTrust/keria/compare/main...lenkan:keria:ruff-check?expand=1#diff-efec556acc809a758da25056e617702341fdc34bd639435343177e138c2e61a5R932

### DIFF
--- a/.github/workflows/python-app-ci.yml
+++ b/.github/workflows/python-app-ci.yml
@@ -28,15 +28,11 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 pytest hio requests
-          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+          pip install -r requirements.txt keria[dev]
       - name: Lint changes
         run: |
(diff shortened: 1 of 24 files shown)


iFergal:
merging as the codecov shouldn't block these code style or formatting changes

PR #386, opened 2025-09-30
enforce and use ruff formatting

lenkan:
For consistent formatting regardless of editor.

Using only the default settings. Tried configuring using single quotes to reduce the diff, but we seem to use both anyway. This makes it consistent.


kentbull:
I like this, and adding it to CI is good, too.

src/keria/app/cli/commands/sig-fix.py
```diff
@@ -118,7 +126,9 @@ def fix(tymth, tock=0.0, **opts):
                         if v.qb64 in prefix_by_next_key_digest:
                             rmids.add(prefix_by_next_key_digest[v.qb64][0].qb64)
 
-                    print(f"\t Proposed smids and rmids updates for {hab.name} - {pre}:")
+                    print(
```
iFergal, line 129: The only real thing I don't like is the line length restriction is making the the code more vertical in places where it doesn't help - like these print statements are just better off on one line. And more vertical height in Python is even harder without brackets like other languages.

But I'm happy with whatever - welcome change overall!

iFergal: I just noticed this 3-line print statement pattern in a few places - I think the ignore would be more annoying in this case. But sure, likely a default for a reason, I'm not pushed!

kentbull:
Let's reformat one more time and I'll approve this.

PR #391, opened 2025-10-21
aiding.py type hints and auto-generated OpenAPI specs

Sotatek-Patrick-Vu:
In this part, we added types hints for all of Ends in aiding.py and auto-generated OpenAPI specs for them


m00sey (approved):
This seems like an overall improvement

Issue #270, opened 2024-07-16
Update POST /identifiers/name/exchanges endpoint to report `recp` field in POST body


pfeairheller:
The POST body for this endpoint currently looks like this (due to the addition of the top level rp field):

```
{
   exn: {
      a: {
         ...,
         i: "RECIPIENT"
      },
      rp: "RECIPIENT",
      ...
   },
   recp: ["RECIPIENT"]
}    
```
This endpoint should be updated to use the field inside the exn and thus removing the body field `recp`.

Arsh-Sandhu:
@pfeairheller  isn't the `recp` field in body required to set the multisig member AID when sending an `exn` message asking multisig members to join an operation?

pfeairheller:
> @pfeairheller isn't the `recp` field in body required to set the multisig member AID when sending an `exn` message asking multisig members to join an operation?

The code should be switched to use the `ri` field in the `exn` message instead.

iFergal:
The only issue here with multi-sig is for each participant you will need to make a Signify <-> KERIA call instead of just 1.

@pfeairheller Does `ri` need to be defined for messages between multi-sig members, and match the local identifier of each recipient? If `ri` is different for each member - so different exns are being sent - it breaks one of the signify tests for `/multisig/rpy` notifications appearing.

lenkan:
@iFergal do you have a link to which test is breaking at which assertion due to that?

iFergal:
@lenkan I'd like to find time to dig into it but it might be because the SAIDs are different. https://github.com/WebOfTrust/signify-ts/actions/runs/10580083902/job/29314065407

```
 FAIL  examples/integration-scripts/multisig-vlei-issuance.test.ts (43.885 s)
  ✕ multisig-vlei-issuance (42765 ms)

  ● multisig-vlei-issuance

    No notifications with route /multisig/rpy

      492 |
      493 |         if (!notes.length) {
    > 494 |             throw new Error(`No notifications with route ${route}`);
          |                   ^
      495 |         }
      496 |
      497 |         return notes;

      at utils/test-util.ts:494:19
      at async retry (utils/retry.ts:34:28)
      at async waitAndMarkNotification (utils/test-util.ts:468:19)
      at async Object.run (multisig-vlei-issuance.test.ts:458:9)
```

iFergal:
[...]

PR #396, opened 2025-11-03
fix: use exn.rp field as recipient for exchange message (#3)

lenkan:
All exchange messages now have the intended recipient in the payload, so the extra body parameter is confusing and could cause issues.

This PR removes the "rec" body parameter that previously specified a list of recipients of an exn message. This list had no correlation with the `rp` field in the exchange message payload.

This PR also removes the ipex multisig handlers, for no other reason that they weren't used in any of the signify integration tests, so to me it was unnecessary complexity. I do not see any reason to keep this feature as we already use the "/multisig/exn" from the client.

See discussions:

- https://github.com/WebOfTrust/signify-ts/pull/267#discussion_r1642546563
- https://github.com/WebOfTrust/keria/issues/270
- https://github.com/WebOfTrust/signify-ts/issues/310
- https://github.com/WebOfTrust/signify-ts/pull/311

### DIFF
--- a/src/keria/app/ipexing.py
+++ b/src/keria/app/ipexing.py
@@ -9,10 +9,8 @@ services and endpoint for IPEX message managements
 import falcon
 from keri import core
 from keri.app import habbing
-from keri.core import eventing, serdering
 from keri.vdr import credentialing
-from keri.peer import exchanging
-
+from keri.core import eventing, serdering
 from keria.core import httping, longrunning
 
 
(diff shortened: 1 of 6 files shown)


kentbull:
@lenkan will you rebase this on main? Also, is it ready for review?

lenkan:
@kentbull Rebased on current main and ready for review. We have been running this code for about 6 months now.

One note on the controversial removal of the multisig IPEX route handlers: no Signify code paths used them at the time. I did not want two different ways of achieving the same thing. The `/multisig/exn` endpoint provides the same capability, and that is what the Signify integration tests use.

src/keria/app/ipexing.py
```diff
@@ -69,18 +67,12 @@ def on_post(req, rep, name):
...
         route = ked["r"]
 
         match route:
             case "/ipex/admit":
-                op = IpexAdmitCollectionEnd.sendAdmit(agent, hab, ked, sigs, rec)
-            case "/multisig/exn":
-                op = IpexAdmitCollectionEnd.sendMultisigExn(
-                    agent, hab, ked, sigs, atc, rec
-                )
+                op = IpexAdmitCollectionEnd.sendAdmit(agent, hab, ked, sigs)
```
iFergal, line 75: Probably worth a refactor now or later to
```
if ked["r"] !== "/ipex/admit":
    raise falcon.HTTPBadRequest(
        description=f"invalid message route {route}"
    )

op = IpexAdmitCollectionEnd.sendAdmit(agent, hab, ked, sigs)
```

iFergal: But really, I have been thinking that an even bigger refactor that would be nice is to drop all IPEX endpoints in favour of just `/identifiers/{name}/exchanges`, and be able to register Decks based on the route, so `agent.grants.append(dict(said=ked["d"], pre=hab.pre))` could be called dynamically.

iFergal (approved):

kentbull:
@lenkan please fix conflicts.

PR #397, opened 2025-11-03
fix: agent config not picked up from parameters

lenkan:
When agent URLs where only provided through parameters (from env variables), the "dt" field was not set, which caused keripy to ignore the configuration. This PR ensure that "dt" field is set.


iFergal (approved):

src/keria/app/agenting.py
```diff
@@ -303,6 +303,7 @@ def _loadConfigForAgent(self, caid):
             config[habName] = {}
 
         config[habName]["curls"] = config[habName].get("curls", [])
+        config[habName]["dt"] = config[habName].get("dt", timestamp)
```
kentbull: I believe the security protocol for reading config is that KERIA, if it wants to be most secure, should reject the config if it doesn't have a "dt" field like how [habbing's Habery.reconfigure](https://github.com/WebOfTrust/keripy/blob/main/src/keri/app/habbing.py#L770) only loads config if `dt` is in conf.

It is more convenient to just add the 'dt' in here if it's missing, though at a bare minimum we should check to see if "dt" is specified and if not either add it in or throw a config error.

lenkan: Thank you for feedback. Can you elaborate on how it is more secure to only load the config if it has a "dt" field?

lenkan: So, just to clarify the flows so we can determine what the difference is:

# Config file

- Create a config file with `curls` and a timestamp dt.
- Start the keria instance
- Some time later, an agent is started, creates and signs `rpy` events `/add/end/role` and `/loc/scheme` with timestamp from config.

# Env variable (with this change)

- Start the keria instance
- Some time later, an agent is started, creates and signs `rpy` events `/add/end/role` and `/loc/scheme` with _current system timestamp_.

Do you agree with this flow? Is the second flow insecure according to the rules that you mentioned?

kentbull (approved):
KRAM dictates timestamps should be checked in order to protect 

There is a security concern with adding a datetime

kentbull:
@lenkan please rebase.

kentbull (approved):

kentbull:
@lenkan approved, please rebase and we'll pull it in.

kentbull (approved):

kentbull:
@lenkan will you rebase this?

PR #401, opened 2025-11-12
feat: index outbound exchange messages

iFergal:
The `ExchangeCueDoer` is hooked up to our 3902 parser so that any successfully parsed and saved exchanges get indexed in the DB afterwards.

This PR hooks up the `Exchanger` to the parsing call in `/identifiers/{name}/exchanges` so that the same happens for outbound exchanges.

This means we can properly find exns we have sent using `/exchanges/query` (this exists in KERIA but not Signify - we are using it on our Signify fork).

@lenkan On `main` this doesn't need to be applied to IPEX as it's using `parser` instead of `hby.psr` but you reverted that in your PR. Once we have aligned on how to handle that there, we can consolidate everything.

### DIFF
--- a/.github/workflows/python-app-ci.yml
+++ b/.github/workflows/python-app-ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, ubuntu-latest]
+        os: [macos-15, ubuntu-latest]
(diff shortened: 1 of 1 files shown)


kentbull:
This also needs to be rebased.

lenkan (approved):

PR #407, opened 2025-12-09
ci(gha): upgrade macos runner to macos-15

iFergal:
`macos-13` runners are now deprecated: https://github.com/actions/runner-images/issues/13046


kentbull (approved):

kentbull:
Looks like this is failing on the macos-13 runner. We should merge it anyway.

PR #418, opened 2026-02-25
feat: persist extern AID metadata for alias-based lookup and rotation

### LINKED ISSUE #415: 500 Error on identifiers.get(alias) for AIDs created with algo: extern
### Version

main, GLEIF-IT/v0.3.0

### Environment

_No response_

### Expected behavior

After creating an AID using an external module(using "signify-ts": "^0.3.0-rc2") (algo: extern), the identifier is retrievable by its alias via the client.identifiers().get(alias) call. The agent returns the correct metadata and maintains a consistent mapping between the alias and the prefix, just as it does for other algorithm types.

### Actual behavior

After successfully creating an AID with algo: extern and confirming that the inception is correctly recorded on the witness network via its prefix, I call client.identifiers().get(alias) to retrieve the identifier's metadata. However, instead of receiving the expected metadata, the call returns a 500 Internal Server Error. This error occurs even though the witness KEL is valid and the prefix lookup works perfectly,

[TEST 1] Lookup by ALIAS: "testV1"
❌ Alias Lookup Failed:
   Status: undefined, Message: "HTTP GET /identifiers/testV1 - 500 Internal Server Error - {\"title\": \"500 Internal Server Error\"}"

[TEST 2] Lookup by PREFIX: "EHJrAyxB4iQylScLdBJR1LntkbzbbeQgc20zqtKoYlqj"
✅ Prefix Lookup Success!
   Sequence Number: 0
   Current Key    : DHHHdqqTEP5aQpmeQ3okXnpV4TYavA89IQh43HphRwTy

### Steps to reproduce

1. Initialize a SignifyClient and inject a custom external module for Ed25519 signing (e.g., using AWS HSM).

2. Create a new AID on the GLEIF Testnet using identifiers().create() with algo: "extern" and a custom extern_type.

3. Verify that the inception event is successfully recorded on the GLEIF Testnet witnesses by querying with the AID prefix.

4. Call client.identifiers().get(alias) using the alias assigned during the creation step.

5. Observe that the agent returns a 500 Internal Server Error instead of the identifier metadata.

yeomjaeseung:
## Description
Resolves #415

This PR completes the `ExternKeeper` implementation as discussed in the issue. Previously, `extern` type AIDs (managed by external modules) lacked a persistence layer in the KERIA database, which led to 500 Internal Server Errors during alias-based lookups and rotations.

By introducing `eprms` storage, this PR ensures that `extern` AIDs achieve functional parity with `salty` and `randy` types, allowing them to be fully managed via aliases within KERIA.

**Key Implementation Details:**
- **Minimal Metadata Only:** To maintain security boundaries, only `extern_type` and `pidx` are persisted. Assumption-free handling of externally managed key material is preserved.
- **Enhanced info() Response:** Extern AIDs now correctly expose their metadata via the `info()` helper, preventing lookup failures in `RemoteManager`.

### `keeping.py`
- Added `ExternPrm` dataclass to store `extern_type` and `pidx`.
- Integrated `eprms` sub-database into `RemoteKeeper` for persistence.
- Implemented `incept`, `rotate`, and `params` methods in `ExternKeeper` to manage the metadata lifecycle.

### `aiding.py`
- Updated `IdentifierResourceEnd.rotate` to support `Algos.extern` metadata updates.
- Updated `AgentResourceEnd.on_put` to allow client-side updates of `extern_type` and `pidx`.

## Verification Results
- **Alias Lookup:** Confirmed `identifiers().get(alias)` returns the correct metadata without 500 errors.
- **Rotation Flow:** Verified that rotating an `extern` AID successfully updates the local DB state.
- **Regression Testing:** Ran full `pytest` suite — all tests passed with no regressions observed.

### DIFF
--- a/src/keria/app/aiding.py
+++ b/src/keria/app/aiding.py
@@ -388,6 +388,19 @@ class AgentResourceEnd:
                         cipher = core.Cipher(qb64=prx)
                         agent.mgr.rb.nxts.put(keys=digers[idx].qb64b, val=cipher)
 
+            elif "extern_type" in val:
+                # extern AIDs keep minimal metadata in KERIA DB (parity with salty/randy).
+                # This enables alias-based lookup/rotation flows to treat extern AIDs as managed identifiers.
+                if (ep := agent.mgr.rb.eprms.get(pre)) is None:
+                    raise ValueError(f"Attempt to update extern for nonexistent pre={pre}.")
+                
+                ep.extern_type = val.get("extern_type", ep.extern_type)
+                if "pidx" in val:
+                    ep.pidx = val["pidx"]
+                    
+                if not agent.mgr.rb.eprms.pin(pre, val=ep):
+                    raise ValueError(f"Unable to update extern prms for pre={pre}.")
+        
         agent.mgr.delete_sxlt()
 
         rep.status = falcon.HTTP_204
@@ -1148,6 +1161,18 @@ class IdentifierResourceEnd:
 
             return op
 
+        elif Algos.extern in body:
+            hab.rotate(serder=serder, sigers=sigers)
+            extern = body[Algos.extern]
+            keeper = agent.mgr.get(Algos.extern)
+
+            try:
+                keeper.rotate(pre=serder.pre, **extern)
+            except ValueError as e:
+                agent.hby.deleteHab(name=name)
+                raise falcon.HTTPInternalServerError(description=f"{e.args[0]}")
+            
+            
         if hab.kever.delpre:
             agent.anchors.append(dict(alias=name, pre=hab.pre, sn=serder.sn))
             op = agent.monitor.submit(
--- a/src/keria/core/keeping.py
+++ b/src/keria/core/keeping.py
@@ -42,7 +42,17 @@ class SaltyPrm:
     def __iter__(self):
         return iter(asdict(self))
 
+@dataclass()
+class ExternPrm:
+    """
+    Extern prefix's parameters for referencing external key management
+    """
+    pidx: int = 0
+    extern_type: str = ""
 
+    def __iter__(self):
+        return iter(asdict(self))
+    
 class RemoteKeeper(dbing.LMDBer):
     """
     RemoteKeeper stores data for Salty or Randy Encrypted edge key generation.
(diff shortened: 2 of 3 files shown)


src/keria/core/keeping.py
```diff
@@ -444,5 +460,36 @@ class ExternKeeper:
...
+
+        if not self.rb.eprms.pin(pre, val=ep):
+            raise ValueError("Unable to rotate extern prms for pre={}.".format(pre))
+
+    def params(self, pre):
+        if (pp := self.rb.pres.get(pre)) is None or pp.algo != Algos.extern:
+            raise ValueError("Attempt to load nonexistent or invalid pre={}.".format(pre))
+        # Default extern params
+        if (ep := self.rb.eprms.get(pre)) is None:
+             return dict(extern=dict(extern_type="", pidx=pp.pidx))
```
iFergal, line 493: For consistency, shouldn't we also `raise ValueError` here as `eprms` should be set in `incept`?

src/keria/app/aiding.py
```diff
@@ -387,6 +387,19 @@ def on_put(self, req, rep, caid):
...
 
+            elif "extern_type" in val:
+                # extern AIDs keep minimal metadata in KERIA DB (parity with salty/randy).
+                # This enables alias-based lookup/rotation flows to treat extern AIDs as managed identifiers.
+                if (ep := agent.mgr.rb.eprms.get(pre)) is None:
+                    raise ValueError(f"Attempt to update extern for nonexistent pre={pre}.")
+                
+                ep.extern_type = val.get("extern_type", ep.extern_type)
+                if "pidx" in val:
+                    ep.pidx = val["pidx"]
```
iFergal, line 399: This makes me wonder if there's actually value in storing any metadata at all with `eprms`, or if it's enough to just populate `pres` and keep it more lightweight.

iFergal (commented):
(sorry for the delay!)

kentbull:
@yeomjaeseung please rebase this on the latest code from main.

m00sey:
As https://github.com/WebOfTrust/keria/pull/429 extracts the interface I'd like to take this into account @yeomjaeseung

yeomjaeseung:
@m00sey 

I agree this should be aligned with #429 rather than handled as a separate persistence-only change.

The intent here was to close the current extern AID lifecycle gap for alias-based lookup and rotation without persisting externally managed key material. At the time, I intentionally kept the implementation minimal since the longer-term direction for the extern interface was still evolving.

Once the extern interface direction in #429 is settled, I’m happy to rebase and adapt this PR accordingly.

PR #421, opened 2026-03-12
bump keria version to 0.2.0, upgrade to keri 1.2.12

kentbull:
Bump to version 0.2.0, upgrade keri to 1.2.12, and update various other dependencies.


daidoji:
Should it be 0.2.0?  On qvi-software GLEIF has it tagged as 0.3.0.  It probably should increment a major version as there are API changes right?  https://github.com/GLEIF-IT/qvi-software?tab=readme-ov-file#gleif-tested-versions-for-technical-dry-runs

kentbull:
Yeah, it would be confusing to have a lower version number than the GLEIF version number. If we're okay with it then I suggest we move to v0.4.0, just to clear things up.

Thoughts @lenkan and @iFergal ?

kentbull:
@daidoji I didn't think there were API changes. If there are then we should do a major version tick to 1.0

daidoji:
I don't know about API changes but does this version communicate with old witnesses?  

Like 0.2.0 worked with keri witnesses on 1.2.6 but I've been having trouble getting those witnesses to work with main.

kentbull:
you know what, let's just eliminate confusion and go with 0.4.0.

Any objections, @daidoji, @iFergal, or @lenkan? I'll close this PR and resubmit a 0.4.0 one.

kentbull:
Succeeded by https://github.com/WebOfTrust/keria/pull/422

Issue #403, opened 2025-11-18
[keria] ERROR eventing .processEscrowA- 2140 Tevery ANC unescrow failed


shanfeng9001:
We have established a lab environment and a production environment.
Currently, during functional verification in the lab environment, we seem to encounter an issue when creating a Registry. This problem prevents us from querying the Registry record after its creation.

The operation steps in our lab environment are as follows:
Created two AIDs (AID-A and AID-B) using the passcode.
AID-A and AID-B performed the "add contact" operation with each other.
AID-A created a Multisig Group, and AID-B joined this Multisig Group.
AID-A and AID-B each executed the "add Enrole" operation, and the MultiSig AID successfully obtained the oobi URL.
AID-A and AID-B respectively performed the Create operation using the same Registry Name and nonce, but an error occurred.

From the Keria console logs, the same error message is continuously output:
[keria] ERROR eventing .processEscrowA- 2140 Tevery ANC unescrow failed: ANC Missing escrowed anchor at dig = EJoWMc8Jmb9Jr4nUsNCDF0NZw6BCdR4TPR1fx8BiPeKA

The above error can be stably reproduced every time. Meanwhile, the multisig groups we created a few months ago can create Registries normally.
Could you please inform us of the ways to troubleshoot this issue? We believe that simply rebuilding the environment may resolve it, but we hope to find a fix while retaining the original test environment.

The relevant code libraries and their versions relied on by the lab environment are listed below:
signify-ts: 0.3.0-rc1
gleif/keria: 0.3.0
gleif/keri: 1.2.8-rc2

iFergal:
I used the versions you provided and it worked for me, so likely there is some small error somewhere in how it's being called. Please provide a snippet of the Signify code which reproduces the error!

shanfeng9001:
Hey @iFergal 

With the same code and identical operational procedures, the multi-signature group we created several months ago can successfully set up the registration center.
We have also conducted multiple tests in this lab environment previously, all of which yielded successful results.
However, in the latest tests, we observed that the creation operation fails on the same day but succeeds when re-executed the next day.
We suspect that clearing the data and rebuilding the service will resolve this issue, but we aim to identify the specific root cause of the problem.
Code snippet:

const client = await getOrCreateClient(props.passcode)
const self = await client.identifiers().get(props.prefix)
const ms = await client.identifiers().get(msAlias)

const resp = await createRegistryMultisig(
	client,
	self,
	ms,
	regName,
	nonce
);

[...]

PR #426, opened 2026-03-23
Fix multisig registry VCP Tevery escrow bug by explicitly processing leader/initiator signature, mirroring `kli multisig join` initiator evt replay

kentbull:
For consistent `kli multisig join` like behavior and to fix `Tevery ANC unescrow` https://github.com/WebOfTrust/keria/issues/403 errors KERIA needs an explicit replay of multisig events from leaders or other group members. This explicit replay ensures that multisig operation joiners will pull in the multisig leader's anchor of events like multisig registry inception (`/multisig/vcp`) and multisig issuance (`/multisig/iss)` events, among others.

Currently KERIA does not explicitly replay this and only accidentally properly stores a multisig anchor in multisig group members of 3 or more members.

This PR fix intermittent errors regarding processing interaction events and their anchors by closing a replay gap between the KLI and KERIA's multisig processing. KLI uses the `JoinDoer` in `join.py` and the `Multiplexor`. 

KERIA’s multisig follower approval path (like `kli multisig join`) is less explicit about event replay from multisig leaders/other members than KLI `JoinDoer` for
- `/multisig/vcp` 
- `/multisig/iss` 
- `/multisig/rev` and 
- `/multisig/rpy`

In KLI, the multisig follower 
- first clones the stored multisig EXN (from the group leader or other member when group size is 3+), 
- parses the initiator’s embedded anchoring event plus its pathed attachments (for example `anc.raw + pathed["anc"]`), and 
- only then parses its own approval of that same event ([keripy `join.py`](https://github.com/WebOfTrust/keripy/blob/main/src/keri/app/cli/commands/multisig/join.py)). 

In KERIA, follower approval :
- goes through `IdentifierResourceEnd.interact(...)`, which processes the locally submitted `ixn` and local `sigs`, 
- and if `"group"` is present, queues that local event into `agent.groups` ([keria `aiding.py`](https://github.com/WebOfTrust/keria/blob/main/src/keria/app/aiding.py)). 
- `GroupRequester` then starts counselor tracking for that local group event ([keria `agenting.py`](https://github.com/WebOfTrust/keria/blob/main/src/keria/app/agenting.py)), 
  - but this generic group path does not itself 
    - fetch the stored EXN, 
    - read `paths["anc"]`, 
    - or replay the initiator’s attached signature stream, meaning the **initiators signature is not processed on the anchoring ixn**, leaving the Tevery event in perpetual escrow until timeout.
      - **This is where the bug occurs, what triggers the seemingly intermittent failure**. 

That makes the current behavior indirect and timing-sensitive: 
- generic `"group"` processing and later `Multiplexor.add(...)` duplicate parsing can sometimes still get to the correct multisig state, 
- 2-of-2 follower flows are especially fragile and can surface as repeated `ANC Missing escrowed anchor` / stalled TEL multisig state convergence ([keripy `grouping.py`](https://github.com/WebOfTrust/keripy/blob/main/src/keri/app/grouping.py)).

### Essentials of this PR

Current state:
- For multisig join the KLI does:
  - `anc = bytearray(aserder.raw) + pathed["anc"]`
  - `self.psr.parseOne(ims=bytes(anc))`
  - then local approval is signed and parsed
- KERIA follower path currently processes:
  - local `ixn`
  - local `sigs`
  - `"group"` metadata
  - but not the stored EXN attachment stream such as `paths["anc"]`
- False positive: Generic `"group"` handling is real, but narrower than KLI replay:
  - `IdentifierResourceEnd.interact(...)` queues the local group event
  - `GroupRequester` starts counselor tracking for that local event
  - neither step clones the stored EXN **from the initiator** or replays pathed attachments, meaning the **initiator signature is ignored**
- `Multiplexor.add(...)` can sometimes rescue multisig state, but only in the case when a later equivalent peer EXN arrives after local approval, which only happens in multisig groups of 3 or more.

Solution:
- Mirror the KLI multisig join follower path by explicitly replaying stored embedded event streams from the initiator before local follower approval. 
  - This ensures the initiator's signature is ALWAYS replayed prior to the local multisig member (follower) event and local signature is processed.

The gap is not “multisig follower approval is absent,” but KERIA does not make the KLI-style stored-attachment initiator signature replay step explicit in follower approval routes.
- This PR's fix: add explicit replay of stored, non-local (initiator) embedded events/attachments (**like signatures**) in the follower approval endpoints for 
  - `/multisig/vcp`
  - `/multisig/iss`
  - `/multisig/rev` and 
  - `/multisig/rpy`

mirroring KLI semantics.

### DIFF
--- a/tests/app/test_credentialing.py
+++ b/tests/app/test_credentialing.py
@@ -7,6 +7,7 @@ Testing credentialing endpoint in the Mark II Agent
 """
 
 import json
+from types import SimpleNamespace
 
 import falcon
 from falcon import testing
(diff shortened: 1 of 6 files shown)


tests/app/test_credentialing.py
```diff
@@ -312,120 +313,124 @@ def test_issue_credential(helpers, seeder):
...
-        data = dict(LEI="254900DA0GOGCFVWB618", dt=dt)
-        creder = proving.credential(
-            issuer=iaid,
-            schema=schema,
-            recipient=recp,
-            data=data,
-            source={},
-            status=registry["regk"],
-        )
+        agent_deeds = doist.enter(doers=[agent])
```
kentbull, line 319: Moved code into the try/except so that the finally block would close out the Doist and release ports once done.

kentbull:
I was incorrect and misunderstood a comment in KERIpy's Multiplexor.add. Closing the PR.

PR #428, opened 2026-03-27
fix: openapi spec field ordering, ACDCs and other misc bugs

iFergal:
The conversion from fielddom to dataclass of ACDC was missing overrides for edges and rules which are always expanded when returned from the API. When compacted variants come in, can be updated to a union with a string but this is more straight forward right now.

`ancatc` is actually always a list of size 1.

Also fixed field orderings that were lost when fields were optional, or inserted via `alts` from fielddom conversion. Now top-level ACDC fields keep their order in the OpenAPI spec.

--

**Edit:** Also realised:
- Registry state was incorrect. Easier to create a new dataclass than use `RegStateRecord` from keripy as that isn't as specific on enums etc.
- Member ends format was wrong (nested another level further)


kentbull (approved):

kentbull (approved):

kentbull:
Please rebase.

PR #431, opened 2026-04-13
Fixes to sig-fix command

### LINKED ISSUE #430: sig-fix command doesn't do anything
### Version

0.4.0

### Environment

linux, 3.13

### Expected behavior

Had to migrate a bunch of legacy wallets from 1.1.33 to keria:0.4.0/keripy:1.2.13.  Ran into https://github.com/WebOfTrust/keria/issues/257 .  Didn't notice at first [that the fix from that issue](https://github.com/WebOfTrust/keria/pull/284) was the "fix-sigs" command.  Thought the `kli migrate` would fix the issue but then found that the command was referenced as the fix for that issue.

Tried to run the command and realized it wasn't doing anything after some debugging.  

sig-fix should run with output (or at least print starting/stopping/failing) behavior.

### Actual behavior

sig-fix exits (successfully `0` return) with no output to console and doesn't seem to run at all (based on some lite debugging).  (Also doesn't fix the issue referenced in #257) 

### Steps to reproduce

Run the command on some wallets produced previously to 1.2.0 as in the referenced issue #257.

daidoji:
Fixes #430

### DIFF
--- a/src/keria/app/cli/commands/sig-fix.py
+++ b/src/keria/app/cli/commands/sig-fix.py
@@ -9,6 +9,7 @@ import argparse
 
 from hio.base import doing
 from keri import help, kering
+from keri.app import directing
 from keri.app import habbing
 from keri.app.cli.common import existing
 from keri.core import serdering, coring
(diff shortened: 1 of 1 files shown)


iFergal (approved):

iFergal:
(I think we can ignore the codecov in this case)

kentbull (approved):

PR #432, opened 2026-04-22
feat: support KERIpy-style /delegate/request EXN sending

kentbull:
We need KERIpy-style `/delegate/request` EXN sending like [line 199](https://github.com/WebOfTrust/keripy/blob/5e5674219c703df0c13612412bb383ce59b42650/src/keri/app/delegating.py#L199) in the Anchorer.partialWitnessEscrow in the `main` branch. This has been a missing feature in KERIA for a long time.

This delegation request EXN is sent from the delegate to the delegator in order for the delegator to get a notification that it can surface to a Signify UX where it shows the edge controller that a delegation request has come through.

In order for single sig delegation to have a KERIpy-style workflow this change is necessary.

I also added protocol, module, class, and function docs to clarify this change.


src/keria/app/delegating.py
```diff
@@ -213,6 +247,93 @@ def processPartialWitnessEscrow(self):
...
+    becomes a normal KERIA notification. Without it the raw delegated event can
+    still be processed eventually, but delegator-facing wallet UX has no
+    actionable request to show.
+    """
+
+    delreq = DelegateRequestHandler(hby=hby, notifier=notifier)
+    exc.addHandler(delreq)
+
+
+class DelegateRequestHandler:
```
iFergal: All of these look identical to the keripy implementation - I think we should import them

docs/protocol.md
```diff
@@ -86,6 +86,80 @@ Once these steps are complete the Signify Client can begin using the rest of the
 Document the steps for retrieving state from the Admin interface and updating.
 
 
+## Delegated Identifier Request Notifications
```
iFergal: I don't think this is the right place for these docs - this doc is for auth between KERIA/Signify and key management

src/keria/app/delegating.py
kentbull: I agree, I'll just import it. Good catch.

docs/protocol.md
kentbull: On second review, I agree. I moved this to the keria_notifications.md doc.

src/keria/app/delegating.py
iFergal: @kentbull I also meant `loadHandlers` and `delegateRequestExn` - I don't see a reason to not import them

docs/keria_notifications.md
```diff
@@ -0,0 +1,86 @@
...
+stream.
+
+## Overview
+Modules leveraging notifications include:
+- delegation
+- multisig identifier creation
+- multisig signing
+- credential issuance and presentation, single or multisig
+- challenge response and verification
+- identifier creation
```
iFergal: Notifications are generally driven by exchange message responses. What notification do we get for identifier creation that isn't multisig (mentioned above)?

docs/keria_notifications.md
```diff
@@ -0,0 +1,86 @@
...
+  "s": "delegate event sequence number",
+  "d": "delegate event SAID"
+}
+```
+
+Wallets should derive the seal from the event instead of
+hard-coding the inception case.
+
+##### Inception
+For delegated inception, `s` is `"0"` and `d` is usually the delegate prefix.
```
iFergal: Remove usually

docs/keria_notifications.md
```diff
@@ -0,0 +1,86 @@
...
+AID, because Signify habitats do not sign inside KERIA. For a group delegated AID, the sender is 
+the local member habitat.
+
+The delegator must already know (have OOBI resolved) the EXN sender's KEL before its exchanger can
+validate the request and create a notification. If the sender is unknown, the inbound EXN will not
+produce a `/delegate/request` notification, even though the raw delegated event may still be sent
+separately.
+
+We should add at the very least an error message in the logs when receiving an approval request from
+an unknown sender, though ideally that would be surfaced to the user in some visible way. Right now
```
iFergal: Surfacing every unknown message would be problematic in production imo

docs/keria_notifications.md
```diff
@@ -0,0 +1,86 @@
...
+##### Rotation
+For delegated rotation, `s` is the rotation sequence number and `d` is the
+rotation event SAID. 
+
+### Sender Verification Precondition
+
+The `/delegate/request` EXN is signed by the sender's habitat selected by KERIA's delegation
+`Anchorer`. For a Signify managed single-sig AID, that sender is normally the delegatee agent/proxy
+AID, because Signify habitats do not sign inside KERIA. For a group delegated AID, the sender is 
+the local member habitat.
```
iFergal: I presume we actually do this because the delegated AID won't be valid until it is approved, so the exn wouldn't surface to the delegator

docs/keria_notifications.md
kentbull: I mistakenly added identifier creation to the list. I will remove it.

docs/keria_notifications.md
kentbull: Agreed. This is now fixed.

docs/keria_notifications.md
kentbull: Fair. Maybe it should be a debug log.

docs/keria_notifications.md
kentbull: Yes, that is correct. A delegated AID cannot sign itself because it is not valid until it is approved. I will add language clarifying this.

src/keria/app/delegating.py
kentbull: This has been resolved. We now use KERIpy's DelegateRequestHandler, loadHandlers, and delegateRequestExn.

docs/keria_notifications.md
```diff
@@ -0,0 +1,216 @@
...
+    - admit: route `/ipex/admit` - Admit the disclosed credential
+    - spurn: route `/ipex/spurn` - Reject either an apply, offer, or agree event.
+- `/challenge/response`: 
+  - Handled by `ChallengeHandler` - does not create a notification, does not need one.
+  - Sends add a signal to Signaler.
+- `/delegate/request`: 
+  - Handled by `DelegateRequestHandler` - creates a notification for the delegator only.
+  - Only one event type, the delegation request.
+- `/oobis`: 
+  - Handled by `OobiRequestHandler`, though is not currently registered. No notification needed for OOBI resolution.
```
iFergal, line 91: If we don't intend on using it in KERIA, I would omit it from the doc entirely to not confuse

kentbull: It seems okay to me in a document dedicated to notifications since it lists something that the underlying library supports. I haven't fully thought through whether `/oobis` would be worth using or not. I included it here for completions sake.

iFergal: This is documentation for KERIA though; I wouldn't bloat it by considering every possible thing we might add from keripy or the documentation won't be focused enough. Anyway, I'll approve as the one line isn't a big concern

iFergal (approved):

docs/keria_notifications.md
kentbull: That's fair. I'll remove this from the docs in my next PR.

PR #435, opened 2026-05-06
feat: SSE broadcaster for agent events

kentbull:
This adds AgentSignals, SseBroadcaster, SseBroadcasterDoer, SignalsStreamEnd , and SseEventIterable to support Signify clients connecting to a server sent events stream from a given Agent to any connected clients for that agent.

This is in preparation for future event driven features including the did:webs integration, among others. The SSE event stream is also a convenient way for clients to switch from long-polling to observing events sent from an agent. Long-polling can still be supported as a fallback, though SSE can be the default. This will significantly help with both bandwidth and performance on highly used Agency servers.

The `SseEventIterable` uses the same pattern as KERIpy's [`QryRpyMailboxIterable`](https://github.com/WebOfTrust/keripy/blob/5e5674219c703df0c13612412bb383ce59b42650/src/keri/app/indirecting.py#L983) to leverage Falcon's `rep.stream` to open an SSE event stream to the client. Wrapping the `SseBroadcaster.subscribe` around the `SseEventIterable` allows HIO to asynchronously service the WSGI's synchronous SSE connection client.

### DIFF
--- a/src/keria/app/streaming.py
+++ b/src/keria/app/streaming.py
@@ -1,0 +1,183 @@
+# -*- encoding: utf-8 -*-
+"""Generic signed agent event streaming helpers for KERIA.
+
+This module owns the reusable transport contract between one KERIA agent and
+its connected edge clients. Topic modules may publish events here, but topic
+modules must not own SSE framing, subscriber fan-out, or KERI ``rpy`` envelope
+signing.
+"""
+
+import json
+import time
+from collections import deque
+
+import falcon
+from hio.base import doing
+from keri.core import eventing
+
+from .. import log_name, ogler
+
+logger = ogler.getLogger(log_name)
+
+
+class SseBroadcaster:
+    """In-memory per-agent SSE broadcaster with independent subscriber queues."""
+
+    def __init__(self):
+        self.subscribers = {}
+        self._index = 0
+
+    def subscribe(self):
+        self._index += 1
+        sid = str(self._index)
+        queue = deque()
+        self.subscribers[sid] = queue
+        return SseEventIterable(self, sid, queue)
+
+    def unsubscribe(self, sid: str):
+        self.subscribers.pop(sid, None)
+
+    def publish(self, event: str, data: dict, event_id: str):
+        payload = json.dumps(data).encode("utf-8")
+        frame = {
+            "id": event_id,
+            "event": event,
+            "data": payload,
+        }
+        for queue in list(self.subscribers.values()):
+            queue.append(frame)
+
+
+class SseEventIterable:
+    """SSE iterable modeled after KERIpy signaling without shared draining."""
+
+    TimeoutSSE = 300  # seconds
+
+    def __init__(self, broadcaster: SseBroadcaster, sid: str, queue, retry=5000):
+        self.broadcaster = broadcaster
+        self.sid = sid
+        self.queue = queue
+        self.retry = retry
+        self.start = None
+        self.end = None
+
+    def __iter__(self):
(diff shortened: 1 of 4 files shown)


kentbull:
See https://github.com/WebOfTrust/signifypy/pull/145 for the SignifyPy client implementation of this.

kentbull:
And see the SignifyTS PR https://github.com/WebOfTrust/signify-ts/pull/376 for the TypeScript client-side implementation of this.

src/keria/app/streaming.py
```diff
@@ -0,0 +1,183 @@
...
+              text/event-stream:
+                schema:
+                  type: string
+                  description: SSE frames with id, retry, event, and JSON data fields.
+        """
+        agent = req.context.agent
+        rep.status = falcon.HTTP_200
+        rep.content_type = "text/event-stream"
+        rep.set_header("Cache-Control", "no-cache")
+        rep.set_header("connection", "close")
```
iFergal, line 182: If this is meant to be long lived, we shouldn't have this header

src/keria/app/streaming.py
```diff
@@ -0,0 +1,183 @@
...
+    sigs = agent.agentHab.sign(ser=rserder.raw)
+    return {"rpy": rserder.ked, "sigs": [siger.qb64 for siger in sigs]}
+
+
+def loadEnds(app):
+    """Register generic signed agent event streaming routes."""
+    app.add_route("/signals/stream", SignalsStreamEnd())
+
+
+class SignalsStreamEnd:
```
iFergal, line 152: I presume the 2 use cases are notifications and long running operation completions. So will this endpoint be phased out in favour of those later? (OK by me)

src/keria/app/streaming.py
```diff
@@ -0,0 +1,183 @@
...
+        {
+            "event": event,
+            "route": route,
+            "payload": payload,
+            "event_id": event_id,
+        }
+    )
+
+
+def signedReplyEnvelope(agent, route: str, payload: dict) -> dict:
```
iFergal, line 138: This probably requires a discussion on a dev call, but does it need to be a reply envelope? Maybe a SPAC message or something makes the most sense, even if unencrypted for now.

src/keria/app/streaming.py
```diff
@@ -0,0 +1,183 @@
...
+        self.broadcaster = broadcaster
+        super().__init__(tock=tock)
+
+    def recur(self, tyme=None, tock=0.0, **opts):
+        while self.cues:
+            cue = self.cues.popleft()
+            try:
+                self.broadcaster.publish(
+                    event=cue["event"],
+                    data=signedReplyEnvelope(
```
iFergal, line 113: Possibly could sign the whole payload rather than the data

src/keria/app/streaming.py
iFergal: Really, we need to be able to replay using the "Last-Event-ID" header when a client reconnects, so I think it might be pragmatic to build this against notifications, as something that is DB backed

src/keria/app/streaming.py
```diff
@@ -0,0 +1,183 @@
...
+            "data": payload,
+        }
+        for queue in list(self.subscribers.values()):
+            queue.append(frame)
+
+
+class SseEventIterable:
+    """SSE iterable modeled after KERIpy signaling without shared draining."""
+
+    TimeoutSSE = 300  # seconds
```
iFergal, line 54: Not an SSE expert but this might be short, and we may want to consider heartbeats to keep the connection open

Issue #316, opened 2024-11-05
Joined member of multisig can't access registries


daviddm:
There’s no way for a new member in a multisig to catch up with all previously created registry in a multisig.

version: Current stable release latest/0.2.0-dev4 (2024-11-05)

I’ve created a test case to highlight that there’s no functionality for a multisig to extend or change their members and continue normal issuance operations.
The test case stops at the registry problem but the same catchup needs to happen with issued credentials from the registry.

# Steps to reproduce
1.	Create multisig with member A and B
2.	Member A and B creates a registry for the multisig
3.	Member A and B double rotates in new member C into the multisig
4.	Member C lists the registries for the multisig

# Expected result
Member C gets a list with registries belonging to the multisig.

# Actual result
Member C gets an empty list of registries for the multisig.

# Reproduction
https://github.com/nordlei/vlei-sandbox/blob/multisig-join-credential/src/issues/multisig-join-registry-problem.test.ts

# Related
https://github.com/WebOfTrust/keria/issues/283

2byrds:
@m00sey and @Arsh-Sandhu maybe you can help us know how this is done in KERIA/keripy? Perhaps there are scripts to update the underlying database? Any inspiration is welcome.

2byrds:
@Arsh-Sandhu will add details for how he did this here <clapping>
This was part of the migration from commandline keripy wallets to KERIA wallets.

Arsh-Sandhu:
To migrate an existing registry of the multisig AID to the wallet of a newly added member, follow these steps:

- From the wallet of an existing member of the multisig, export any or all `issued` credentials of the multisig AID in CESR format. In Signify, you can use the `Get` method of the `Credentials` class and pass arg `includeCESR=true`.
- Send a `PUT` request to `http://<keria-agent-address>:<keria-agent-port>/` with the headers `CESR-DESTINATION: <local AID of the newly added member in the multisig AID>` and `Content-Type: application/octet-stream`, and include the exported credential data in the body. For example:
`curl -X PUT http://<keria-agent-address>:<keria-agent-port>/ --data-binary @credentials.cesr -H "CESR-DESTINATION: <local AID of newly added member into multisig AID>" -H "Content-Type: application/octet-stream"`
- This will store the exported credential(s) and the registry in the new member's wallet.
- At this point, the registry does not have a name. 
[...]

PR #439, opened 2026-06-04
Drain imported CESR before registry rename

kentbull:
References WebOfTrust/keria#316.

This drains any CESR bytes queued by an agent import before the registry rename endpoint aliases an imported registry as local. In the late-join multisig flow, the credential CESR import can queue TEL material and the immediate registry rename can otherwise cause the queued issue event to be parsed under local registry handling.

The change keeps the import/rename ordering deterministic without changing the public API.

### DIFF
--- a/src/keria/app/credentialing.py
+++ b/src/keria/app/credentialing.py
@@ -575,6 +575,10 @@ class RegistryResourceEnd:
         if "name" not in body:
             raise falcon.HTTPBadRequest(description="'name' is required in body")
 
+        # Drain any just-imported CESR before aliasing an imported registry as local.
+        if agent.parser.ims:
+            agent.parser.parse()
+
         name = body["name"]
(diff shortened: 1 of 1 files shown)


kentbull:
Related PRs for the same WebOfTrust/keria#316 late-join multisig registry work:

- WebOfTrust/signifypy#149
- WebOfTrust/signify-ts#388

iFergal:
This is only really possible in a script race condition, and the registry rename trick is a workaround. imo we should just add a delay in the script and keep this endpoint clean.

Hopefully, moving to registrars will help resolve this issue in a more permanent way that also works outside scripts.

kentbull:
> Hopefully, moving to registrars will help resolve this issue in a more permanent way that also works outside scripts.

Yeah, registrars is the right idea. I agree with you on the delay. I'll close this PR.

PR #441, opened 2026-06-10
chore: fix depends type in some operation metadata

Sotatek-Patrick-Vu:
Adds a common `KelOperation` which also helps type generation in Java.

### DIFF
--- a/src/keria/app/specing.py
+++ b/src/keria/app/specing.py
@@ -881,18 +881,6 @@ class AgentSpecResource:
                 {"$ref": "#/components/schemas/ACDC_V_2"},
             ]
         }
-        self.spec.components.schemas["CredentialOperationMetadata"]["properties"][
-            "depends"
-        ] = {
-            "oneOf": [
-                {"$ref": "#/components/schemas/ROT_V_1"},
-                {"$ref": "#/components/schemas/ROT_V_2"},
-                {"$ref": "#/components/schemas/DRT_V_1"},
-                {"$ref": "#/components/schemas/DRT_V_2"},
-                {"$ref": "#/components/schemas/IXN_V_1"},
-                {"$ref": "#/components/schemas/IXN_V_2"},
-            ]
-        }
         self.spec.components.schema(
             "CredentialOperationResponse",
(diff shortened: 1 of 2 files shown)


src/keria/app/specing.py
```diff
@@ -1002,6 +984,22 @@ def __init__(self, app, title, version="1.0.1", openapi_version="3.1.0"):
             ]
         }
 
+        self.spec.components.schemas["KelOperation"] = {
+            "oneOf": [
+                {"$ref": "#/components/schemas/GroupOperation"},
+                {"$ref": "#/components/schemas/WitnessOperation"},
+                {"$ref": "#/components/schemas/DoneOperation"},
+                {"$ref": "#/components/schemas/DelegationOperation"},
+                {"$ref": "#/components/schemas/SubmitOperation"},
```
iFergal: I think we can remove `SubmitOperation` from this list as this is for re-submitting events to be witnessed, and never as a direct result of creating a registry, issuing an ACDC etc.

src/keria/core/optypes.py
```diff
@@ -382,48 +380,58 @@ class FailedDelegatorOperation(BaseDelegatorOperation, FailedOperation):
...
+SubmitOperation = Union[
+    PendingSubmitOperation, CompletedSubmitOperation, FailedSubmitOperation
+]
+
+KelOperation = Union[
+    "GroupOperation",
+    "WitnessOperation",
+    "DoneOperation",
+    "DelegationOperation",
+    "SubmitOperation",
```
iFergal: Same here

(Note to others - most of the diff in this file is moving the ordering of the delegator classes and submit classes)

Sotatek-Patrick-Vu: Yep, because I added SubmitOperation then I have to re-older the dataclass, but I can revert it after remove SubmitOperation from the union list

iFergal (approved):

kentbull (approved):

kentbull:
@Sotatek-Patrick-Vu please rebase and then I will merge.

PR #445, opened 2026-06-27
feat: suppress redundant IPEX initiator notifications

kentbull:
IPEX operation initiators do not need notifications of events they create as it is redundant. Suppressing notifications for IPEX operations like this, for initiators, is similar to how Multiplexor in KERIpy suppresses notifications for multisig operation initiators.

https://github.com/WebOfTrust/signify-ts/pull/399/changes can be moved from draft to PR once this is merged.

### DIFF
--- a/src/keria/app/ipexing.py
+++ b/src/keria/app/ipexing.py
@@ -12,8 +12,41 @@ from keri.app import habbing
 from keri.core import eventing, serdering
 from keri.vdr import credentialing
 from keri.peer import exchanging
+from keri.vc import protocoling
 
 from keria.core import httping, longrunning
+
+
+IPEX_ROUTES = (
+    "/ipex/apply",
+    "/ipex/offer",
+    "/ipex/agree",
+    "/ipex/grant",
+    "/ipex/admit",
+    "/ipex/spurn",
+)
+
+
+class IpexNotificationHandler(protocoling.IpexHandler):
+    """IPEX handler that suppresses notifications for locally originated EXNs."""
+
+    def handle(self, serder, attachments=None):
+        # Checking if the message sender AID (serder.pre) is local and, if so, skip making a
+        # notification. This prevents a redundant notification appearing for an IPEX operation
+        # initiator.
+        if serder.pre in self.hby.habs:
+            return
+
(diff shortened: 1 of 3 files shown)


src/keria/app/ipexing.py
```diff
@@ -12,10 +12,43 @@
...
+
+class IpexNotificationHandler(protocoling.IpexHandler):
+    """IPEX handler that suppresses notifications for locally originated EXNs."""
+
+    def handle(self, serder, attachments=None):
+        # Checking if the message sender AID (serder.pre) is local and, if so, skip making a
+        # notification. This prevents a redundant notification appearing for an IPEX operation
+        # initiator.
+        if serder.pre in self.hby.habs:
+            return
```
iFergal, line 38: If we want to do this, we should probably just avoid calling `handle` so it will apply to all exns?

There is a nice advantage of keeping the extra notification that we are using in Veridian Wallet which I think is worth considering though.

In a threshold multi-sig, any members who don't sign get a notification that the exn was fully signed automatically, which is useful for UX.

kentbull: > If we want to do this, we should probably just avoid calling handle so it will apply to all exns?

This would require subclassing KERIpy's `Exchanger` since that is what calls `.handle()` on all registered message handlers. I have not yet fully thought through whether this makes sense for all exns, yet in abstract it seems okay. It doesn't seem the initiator of a challenge response, OOBI exn resolution request, IPEX, or multisig operation needs to be notified of their own operation. Maybe we could just avoid calling `handle` like you say.

> There is a nice advantage of keeping the extra notification that we are using in Veridian Wallet which I think is worth considering though.
> In a threshold multi-sig, any members who don't sign get a notification that the exn was fully signed automatically, which is useful for UX.

Does this imply that you are having a non-signer multisig member initiate an event?

iFergal: > Does this imply that you are having a non-signer multisig member initiate an event?

No. For example, for a 2 of 3 multi-sig that receives a presentation request: if Alice responds with a presentation, and Bob also signs, Carol will get a notification of the presentation which indicates that enough signatures were collected and that she doesn't need to sign.

PR #447, opened 2026-07-02
fix: drop agent EID suffix on multisig OOBI by default

kentbull:
This has been a constant source of confusion for years and should return the intuitive result by default, the multisig Agent OOBI with no agent EID suffix.

The includeEid URL parameter allows using the old behavior, if desired.

SignifyTS PR: https://github.com/WebOfTrust/signify-ts/pull/400
SignifyPy PR: https://github.com/WebOfTrust/signifypy/pull/150

### DIFF
--- a/src/keria/app/agenting.py
+++ b/src/keria/app/agenting.py
@@ -40,6 +40,7 @@ from keri.app import (
     connecting,
     grouping,
 )
+from keri.app import delegating as kdelegating
 from keri.app.grouping import Counselor
 from keri.app.keeping import Algos
 from keri.core import coring, parsing, eventing, routing, serdering
@@ -66,7 +67,7 @@ from ..core.httping import falconApp, createHttpServer
 from ..peer import exchanging as keriaexchanging
 from .specing import AgentSpecResource
 from ..core import authing, longrunning, httping
-from ..core.authing import Authenticater
+from ..core.authing import SignedHeaderAuthenticator
 from ..core.keeping import RemoteManager
 from ..db import basing
 from .credentialing import (
@@ -725,6 +726,7 @@ class Agent(doing.DoDoer):
         handlers = [challengeHandler]
         self.exc = exchanging.Exchanger(hby=hby, handlers=handlers)
         grouping.loadHandlers(exc=self.exc, mux=self.mux)
+        kdelegating.loadHandlers(hby=self.hby, exc=self.exc, notifier=self.notifier)
         protocoling.loadHandlers(hby=self.hby, exc=self.exc, notifier=self.notifier)
         self.submitter = Submitter(
             hby=hby, submits=self.submits, witRec=self.witSubmitDoer
@@ -941,15 +943,13 @@ def createAdminServerDoer(config: KERIAServerConfig, agency: Agency):
     Returns the Doer and the Falcon app so the HTTP app can use it for OpenAPI docs.
     """
     # Create Authenticater for verifying signatures on all requests
-    authn = Authenticater(agency=agency)
+    authn = SignedHeaderAuthenticator(agency=agency)
 
-    adminApp = falconApp(config.logRequests)
+    adminApp = falconApp(config.logRequests, request_type=authing.ModifiableRequest)
     if config.cors:
         adminApp.add_middleware(middleware=httping.HandleCORS())
     adminApp.add_middleware(
-        authing.SignatureValidationComponent(
-            agency=agency, authn=authn, allowed=["/agent"]
-        )
+        authing.AuthenticationMiddleware(agency=agency, authn=authn, allowed=["/agent"])
     )
     adminApp.req_options.media_handlers.update(media.Handlers())
     adminApp.resp_options.media_handlers.update(media.Handlers())
@@ -1021,8 +1021,8 @@ def setupDoers(agency: Agency, config: KERIAServerConfig, temp=False, cf=None):
     """
     Sets up the HIO coroutines the KERIA agent server is composed of including three HTTP servers for a KERIA agent server:
     1. Boot server for bootstrapping agents. Signify calls this with a signed inception event.
-    2. Admin server for administrative tasks like creating agents.
-    3. HTTP server for all other agent operations.
+    2. Admin server for any Signify client related actions after bootstrapping.
+    3. HTTP server for all other external agents send KERI events or messages for interactions.
 
     Parameters:
         config (KERIAServerConfig): Configuration for the KERIA server.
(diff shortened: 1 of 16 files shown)


src/keria/app/agenting.py
```diff
@@ -2110,15 +2111,18 @@ def on_get(req, rep, alias):
...
-                url = (
-                    urls[kering.Schemes.http]
-                    if kering.Schemes.http in urls
-                    else urls[kering.Schemes.https]
-                )
-                up = urlparse(url)
-                oobis.append(urljoin(up.geturl(), f"/oobi/{hab.pre}/agent/{eid}"))
-                res["oobis"] = oobis
+            # Outer for loops over multi-valued mict that could have multiple "agent" dict values
+            for eurls in roleUrls.naball(kering.Roles.agent):
```
iFergal: Is this endpoint actually used by Signify, or just `/identifiers/{name}/oobis`? If not I'd be in favour of removing it and keeping one.

src/keria/app/aiding.py
```diff
@@ -1420,22 +1439,23 @@ def on_get(req, rep, name):
...
+                            if kering.Schemes.http in murl:
+                                urls.extend(murl.naball(kering.Schemes.http))
+                            if kering.Schemes.https in murl:
+                                urls.extend(murl.naball(kering.Schemes.https))
+                            for url in urls:
+                                oobi = agentOobiUrl(
+                                    hab, url, agent, includeEid=includeEid
+                                )
+                                if oobi not in oobis:
+                                    oobis.append(oobi)
```
iFergal: A bigger, more breaking change but I think it might be worth considering moving from returning URLs to returning objects and letting the client build the URL.

For example, if we returned objects like `{"eid": "EJsP...", "scheme": "http", "url": "http://keria:3902/"}` all of the information would be available, and the client decides.

With a default that we build URLs _without_ the `eid` suffix, as it's not really adding any value beyond specific cases where you want the filtered view (unsure on what those cases are).

src/keria/app/agenting.py
kentbull: Yes, it's not used. I'll remove it and move any valuable ideas to `/identifiers/{name}/oobis`

src/keria/app/aiding.py
kentbull: > A bigger, more breaking change but I think it might be worth considering moving from returning URLs to returning objects and letting the client build the URL.

I really like that. Let's plan on doing that when we upgrade to 1.0.0. This would allow us to return a rich object and let the client make the decision. Good idea. Please make an issue about this and let's add it to the prioritized roadmap.

iFergal: Note that this change is a breaking change, so if we do that, it's 2 breaking changes to this API.

For example, Veridian Wallet right now will search for `/agent/` which won't exist on this API (just `/agent`). Easy fix, but there may be many clients. So if we will put a breaking change, maybe it should be just done once correctly. What do you think?

kentbull: Well, maybe we could leave the old API there and make a new endpoint, or a query parameter, that allows returning the object. We could maintain the old API indefinitely and just deprecate it. Then it doesn't have to be breaking.

A breaking change to a core component like this in such a small ecosystem, on second thought, is risky. We could mitigate that risk by allowing an optional upgrade path with default support for the old one.

kentbull: > Note that this change is a breaking change

I believe that returning a multisig OOBI with an agent EID on the end was a bug, and while stripping the EID by default could be considered a breaking change if someone relied on that, then I am not aware of anyone who actually relied on it. 

Technically, yes, it is breaking, yet only for multisig, and only for people that, for whatever reason, needed the filtered view of the single-member's view of the multisig KEL + end roles and loc schemes.

This changes nothing for single sig identifiers and leaves the Agent EID on the end of their OOBIs, so, in my mind, it seems okay to merge.

kentbull: > For example, Veridian Wallet right now will search for /agent/ which won't exist on this API (just /agent). Easy fix, but there may be many clients.

Oh, I see what you're saying. Darn. I do remember you saying Veridian does not yet do a lot with multisig. Does Veridian do the `/agent/` search on multisig identifiers?

> So if we will put a breaking change, maybe it should be just done once correctly. What do you think?

Yeah, if we're going to break it, let's do it once. I would still favor having the object being returned be an optional upgrade since we've got a lot of QVI clients at GLEIF that we don't want to break by changing this API.

And that makes me think that maybe we have to support the old `/agent/` version indefinitely...hmm.

Or maybe we should just break it once and include an upgrade guide...

I favor breaking it once and forcing people to upgrade so we stop getting so many inadvertent incorrect usages of the API.

iFergal: > Oh, I see what you're saying. Darn. I do remember you saying Veridian does not yet do a lot with multisig. Does Veridian do the /agent/ search on multisig identifiers?

Veridian has had multi-sig since the start. I'm saying Veridian drops the EID suffix using `.indexOf("/agent/");`, and it's possible that there are other QVIs which also do this.

> And that makes me think that maybe we have to support the old /agent/ version indefinitely...hmm.

While it's worth trying to avoid breaking changes, KERIA is still not even at a 1.0.0 release so one way or another there will be breaking changes. Breaking in a small ecosystem is better than a large one. I also think the Signify APIs are very inconsistent (I know @lenkan thought the same) and need re-working with breaking changes.

We could alternatively maintain a release branch for this pre 1.0.0 version, and put some effort into a more polished API on main. Perhaps something to discuss next Tuesday with a wider group.

kentbull: I just realized last night we could also add some intelligence to our OOBI resolution in KERIA that even if a multisig OOBI is resolve we check the AID in the OOBI (not the EID) and if it is a group AID then we would just resolve it as a normal group AID. This would be okay because there are literally zero use cases for resolving the filtered view.

We could add a query param to resolve the filtered view if someone complains and asks for it. Then we can avoid breaking backwards compatibility with any client.

I'll look into this today.

So, this PR could still change the default, yet the old behavior would also be supported.

iFergal: Hmm, we could and it would work. It feels like hidden behaviour though and might end up being confusing. I think I would accept it as a non-breaking compromise for now, and move to my suggestion later. Perhaps we can discuss tomorrow

kentbull: Yeah, that's my worry, that it is hidden behavior. Compromise is the right description.

Okay, sure, let's discuss tomorrow.

Issue #213, opened 2024-03-18
Provide a way to rollback partially signed multisig events 


lenkan:
Consider a group X consisting of Member A and Member B. If both member A and member B create an event for Group X at the same time.

1. Member A issue credential C1  (sequence number: 1)
1. Member B issue credential C2 (sequence number: 1 because they are still unaware of the event created by Member A)
1. Member A sign credential issuance event for C2 (what is the sequence here ?)
1. Member B sign credential issuance event for C1 (sequence ?)

After step 2, member A and member B has diverging KELs for Group X. We need a way to rollback or resolve this conflict.

From question on discord:

> Phil Feairheller — Today at 2:10 PM
> > In the command line of KERIpy there is a multisig rollback command for deleting partially signed events at the tip of your KEL. I don’t think anyone has added that functionality to KERIA.

2byrds:
See https://github.com/WebOfTrust/signify-ts/pull/286 which addresses the 'catchup' for member 2 (they have the exn, notification).

2byrds:
> Would a general design approach to consider for resolving various multisig group ceremonies that could get stuck be to introduce a time-to-live concept? In such an approach where these ceremonies are stuck would get reset (expired) because the initial request would expire if not fully committed after the default TTL for that type of request or an explicit TTL in the request.

@edeykholt a lot of these issues are with one member getting out-of-sync with the group ops (the group ops succeeded). in terms of the out-of-sync member, i think the rollback i'm implementing could eventually be part of the 'time-to-live' that you mention. Perhaps it should be configurable so that the member can configure their expectation for the event, something like 'Lets issue this credential over the next 60 minutes, but if it doesn't finish by then, rollback this operation so we can try again later'.

edeykholt:
@2byrds Seems like a good step forward.

PR #449, opened 2026-07-14
fix: prevent stale group credential acceptance blocking issuance

lenkan:
## Problem

A stale multisig credential-issuance acceptance could poison a member while another issuance was in flight:

1. In a 2-of-2 group, M1 issues credential 1 and M2 accepts it successfully.
2. M1 starts credential 2, leaving M2's `/multisig/iss` notification unread.
3. M2 accepts credential 1's old exchange again. The wallet creates a fresh KERIA issue request from credential 1's completed ACDC; it does **not** replay the original HTTP request, KEL, TEL, or signatures.
4. M2 then accepts credential 2.

Before this change, step 3 could return HTTP 500 and leave M2 unable to finish or read credential 2. M1 could also remain blocked waiting for M2's contribution.

Related to #213.

## Fix

Treat an already-completed group credential acceptance as idempotent by reusing its persisted operation state, without submitting a conflicting KEL or TEL event.

### DIFF
--- a/src/keria/app/credentialing.py
+++ b/src/keria/app/credentialing.py
@@ -10,7 +10,7 @@ import json
 from dataclasses import asdict, dataclass, field
 
 import falcon
-from keri import kering, help
+from keri import core, help, kering
 from keri.app import signing
 from keri.app.habbing import SignifyGroupHab
 from keri.core import coring, scheming, serdering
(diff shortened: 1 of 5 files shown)


lenkan:
@iFergal @kentbull is this something you have experienced?

iFergal:
@lenkan No, but I am in favour of making the APIs idempotent

PR #454, opened 2026-07-24
feat: add generic /verify endpoint

Sotatek-Patrick-Vu:
The /credentials/verify was previously used to verify a credential from Signify for use cases where ACDCs are published somewhere and indexed, such as a blockchain, rather than direct IPEX engagements.

We would like to supersede the endpoint with a more generic /verify endpoint which accepts a serder + attachment and returns the relevant long running operation - ACDC + TEL events supported for now.

Draft PR for now; looking for feedback on general approach.

### DIFF
--- a/src/keria/core/longrunning.py
+++ b/src/keria/core/longrunning.py
@@ -26,7 +26,7 @@ from keria.app import delegating
 Typeage = namedtuple(
     "Tierage",
     "oobi witness delegation group query registry credential endrole "  # type: ignore[name-match]
-    "locscheme challenge exchange submit done",
+    "locscheme challenge exchange submit verifyCredential done",
 )
 
 OpTypes = Typeage(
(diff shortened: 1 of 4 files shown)


src/keria/core/longrunning.py
```diff
@@ -477,7 +477,7 @@ def status(self, op):
                 )
 
             ced = op.metadata["ced"]
-            if self.credentialer.complete(ced["d"]):
+            if self.credentialer.rgy.reger.saved.get(keys=ced["d"]) is not None:
```
iFergal, line 480: `credentialer.complete` only works for credentials we issue - looking at the escrows, there isn't really much difference right now compared to `reger.saved` which would allow this op type to be used for general verification of credentials too.

Not sure if this would change when we have deployed registrars is the only thing. Happy to split to two if we think that'd be better.

iFergal (commented):
@kentbull @lenkan Looking for general feedback on the approach, hence draft, but it's otherwise a working PR.

PR #457, opened 2026-08-15
Finalize ESSR HTTP envelopes after Falcon responses

kentbull:
## Summary

- parse decrypted ESSR request envelopes into a complete, validated WSGI 1.0 environment
- default HTTPS authorities to port 443, support bracketed IPv6 authorities, and preserve request body bytes exactly
- move ESSR response protection behind Falcon's WSGI finalization so media, cookies, status semantics, and `Response.stream` are handled correctly
- buffer only bounded finite response streams, which fixes the existing contact-image download path, and fail fast without consuming unknown-length streams such as SSE
- document that the inner reader is an ESSR-specific envelope parser, not a general HTTP parser

## Protocol boundaries

ESSR v1 seals one complete inner HTTP message, so it cannot progressively transport a live stream. This change supports the existing bounded contact-image stream because it has a known `Content-Length`; unknown-length streams produce an encrypted inner 501 response. Future SSE endpoints should use signed-header mode unless a framed streaming-encryption protocol is specified.

## Validation

- `make test` (105 passed)
- `uv run pytest tests/core/test_authing.py tests/core/test_httping.py -q` (27 passed)
- `uv run ruff check src tests`
- `uv run ruff format --check src tests`
- `uv run --with myst-parser sphinx-build -b html docs /tmp/keria-sphinx.Jkowbv`

The normal Sphinx build succeeds. A warnings-as-errors build remains blocked by pre-existing repository documentation debt (139 autodoc/docstring warnings) and by the undeclared `myst_parser` documentation dependency; neither is introduced by this change.


kentbull:
wrong branch, meant for 1.2.14

Issue #438, opened 2026-05-21
AttributeError: 'NoneType' object has no attribute 'verfers' for "sig-fix"


daidoji:
### Version

0.4.1

### Environment

linux, 3.14

### Expected behavior

sig-fix command should work for all wallets.

### Actual behavior

While attempting to run sig-fix command on very old wallets we got the following stack trace.

Traceback (most recent call last):
  File "/keria/.venv/bin/keria", line 10, in <module>
    sys.exit(main())
             ~~~~^^
  File "/keria/.venv/lib/python3.13/site-packages/keria/app/cli/keria.py", line 27, in main
    raise ex
  File "/keria/.venv/lib/python3.13/site-packages/keria/app/cli/keria.py", line 22, in main
    args.handler(args)
    ~~~~~~~~~~~~^^^^^^
  File "/keria/.venv/lib/python3.13/site-packages/keria/app/cli/commands/sig-fix.py", line 25, in <lambda>
    parser.set_defaults(handler=lambda args: handler(args), transferable=True)
                                             ~~~~~~~^^^^^^
  File "/keria/.venv/lib/python3.13/site-packages/keria/app/cli/commands/sig-fix.py", line 40, in handler
    return directing.runController([doing.doify(fix, **kwa)], expire=0.0)
           ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/keria/.venv/lib/python3.13/site-packages/keri/app/directing.py", line 678, in runController
    doist.do(doers=doers)
    ~~~~~~~~^^^^^^^^^^^^^
  File "/keria/.venv/lib/python3.13/site-packages/hio/base/doing.py", line 156, in do
    self.recur()  # increments .tyme runs recur context
    ~~~~~~~~~~^^
  File "/keria/.venv/lib/python3.13/site-packages/hio/base/doing.py", line 275, in recur
    tock = dog.send(self.tyme)  # yielded tock == 0.0 means re-run asap
  File "/keria/.venv/lib/python3.13/site-packages/keria/app/cli/commands/sig-fix.py", line 125, in fix
    for v in hab.kever.verfers:
             ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'verfers'

### Steps to reproduce

Get a really old wallet (probably from before verfers existed) and try to run the sig-fix command against it.

PR #458, opened 2026-08-27
Added some error handling to sig-fix.

daidoji:
Habs and caids that end up in states for which the command won't work now won't error out for the whole run.

Also added some logging statements to the end showing the malformed AIDs and how many AIDs were processed by the script.

Doesn't fix but bypasses #438 and an additional bug where mid not existing in the past for some habs caused the existingHby open to fail.


daidoji:
sig-fix wasn't really tested before and so I didn't really add any tests for it now.

kentbull (approved):

kentbull:
@daidoji please update this with the base branch.

PR #464, opened 2026-09-14
upgrade and prep: to KERIpy 1.2.14 and minimal strict tocks compat

kentbull:
# PR Core: 

**keri 1.2.14 upgrade including minimal tocks compatibility with newly strict upstream KERIpy tock config validation**

- Upgrade KERIA to `keri==1.2.14` from PyPI. 
- Keep `hio==0.6.14`: the published KERIpy release requires `hio>=0.6.14`, rather than a newer exact HIO version. The lockfile changes only KERIpy and its artifact metadata.

KERIpy 1.2.14 validates tock configuration **strictly**, so a dependency-only upgrade rejects existing KERIA configuration such as `{"tocks": {"escrower": 1.0}}`, thus the need for the tock config work in this PR. 

### Backwards compatibility for legacy flat tocks

Reserves the twelve existing KERIA Agent keys, now referred to as "legacy" tock config keys, when resolving KERIpy tocks before Agent Habery creation and reopening. 

Preserves the original, legacy configuration for Agent consumers, while retaining KERIpy validation for new tock config and `cli -> env var -> config file` based override precedence. 

The test helper uses the same legacy config loading path.

## Overall Tocks Plan

This is the prerequisite dependency PR for the planned KERIA tock configuration work coming after this PR merges. It does not yet introduce the KERIA registry, namespace migration to the new `"tocks.signify"` subsection of the config file, deprecation warnings for using legacy tocks, runtime scheduling changes, or new performance defaults.


kentbull:
If possible will one of @daidoji @Arsh-Sandhu also review this PR? This is prep work for me to add the 100% tock configurability to KERIA, which will be the next PR.

kentbull:
I should say, this PR also pulls in the ESSR changes into KERIA that the Veridian team graciously contributed upstream to KERIpy. There are a number of things in the recently released `v1.2.14` that you can see in the [changelog](https://github.com/WebOfTrust/keripy/blob/v1.2.14/ref/ChangeLog.md#1214-unreleased). It says "unreleased" in the changelog yet that just needs to be updated because it was released.

src/keria/app/agenting.py
```diff
@@ -164,6 +165,32 @@ def readConfigFile(configDir: str, configFile: str, temp=False):
     )
 
 
+def keriTocks(cf):
```
iFergal, line 168: For my understanding, if the output of `keriTocks` is only passed to keripy classes like `habbing.Habery`, why do we need to take the ones defined in KERIA into account? As those are just used in agenting.py Doers

daidoji (approved):

src/keria/app/agenting.py
kentbull: Because unknown config keys [now raise](https://github.com/WebOfTrust/keripy/blob/v1.2.15/src/keri/app/tocking.py#L226) as of KERIpy v1.2.14 introducing strict config validation in a move towards avoiding invalid/garbled config values causing problems. The strict tock config schema is stored in [`TOCKS`](https://github.com/WebOfTrust/keripy/blob/v1.2.15/src/keri/app/tocking.py#L49) and is incrementally built as a tock registry as each tock is defined in [`tocking.py`](https://github.com/WebOfTrust/keripy/blob/v1.2.15/src/keri/app/tocking.py).

So, the strict rules now require a specific config schema, meaning
```json
{
  "tocks": {
      "initer": 0.0,
      "escrower": 1.0
  }
}
```
under keripy v1.2.14 would raise a config error on `tocks.initer` and `tocks.escrower` unless they were passed in `reserved` to `tocking.resolveTocks`.

The following would not raise once the next PR in the tocks sequence is added:
```json
{
  "tocks": {
    "signify": {
      "initer": 0.0,
      "escrower": 1.0
    }
  }
}
```
This would not raise because the plan is for KERIA-specific tocks to be in a `TOCKS` like object, probably `SIGNIFY_TOCKS` or `KERIA_TOCKS`, defined under the `tocks.signify` sub object, as a tock registry for KERIA tocks.

The purpose of `reserved` in `tocking.resolveTocks` is both to support this kind of backwards compatibility while also allowing coarse grained config props to cascade down to multiple leaf/fine-grained config props.

iFergal: OK, makes more sense in the context of the next PR then!

iFergal (approved):

Issue #258, opened 2024-06-27
Add support for configuration file entries for Agent Doer tocks


pfeairheller:
Each Doer for the Agents should have their own tock values read from a configuration file along with a sane default for the processing done by that Doer.

This will allow for installation specific tuning of background processing for KERIA

rubelhassan:
@pfeairheller please assign me.

kentbull:
Working on this here: https://github.com/GLEIF-IT/keria/pull/6

Pulling in the same tock configuration approach applied to the v1.2.x line of KERIpy releases.

kentbull:
Once KERIpy v1.2.14 has been released in PyPi then I'll pull the supported tock configuration PR from the GLEIF-IT fork to this upstream repo.

kentbull:
https://github.com/WebOfTrust/keria/pull/465 fully addresses this need.

PR #465, opened 2026-09-15
Add KERIA tock configuration with legacy compatibility

kentbull:
## Summary

KERIA previously read twelve flat Agent tock settings, but lacked consistent environment overrides, validation, and propagation into child tasks. This PR centralizes scheduling configuration in `keria.app.scheduling`, reuses KERIpy's resolver, and threads configured tock cadences through recurring Agency/Agent-managed work. It builds on #464 toward #258 and the 0.4.1 release.

Registered work defaults to `0.03125` seconds, one cycle of KERIA's existing 32 Hz scheduler. Explicit `0.0` remains supported for next-cycle scheduling. The only slower default is `releaser` at `60.0` seconds: scanning all open Agents for the default 24-hour inactivity timeout does not require a scan every scheduler cycle. The root scheduler cadence is unchanged.

## Configuration and compatibility

- Expose **19 KERIA settings** under `tocks.signify` and **29 KERIpy settings** directly under `tocks`. Apply KERIA's one-cycle defaults through a copied core registry without changing KERIpy's process-global defaults.
- Preserve the twelve existing Agent settings and add `submitter`, `bootServer`, `adminServer`, `httpServer`, `shutdown`, `signalExpiry`, and `releaser`.
- Accept legacy flat KERIA keys with a migration warning. For example, `{"tocks":{"escrower":1.0}}` remains valid; the canonical form is `{"tocks":{"signify":{"escrower":1.0}}}`. Removal is deferred to a future announced release.
- Support specific environment overrides such as `KERIA_ESCROWER_TOCK` and the group override `KERIA_AGENT_TOCK`. The Agent group covers its 13 work settings and excludes servers, shutdown, signal expiry, and release scans. Core aliases remain within the KERIpy namespace.
- Preserve precedence: **specific environment > group environment > specific config > group config > default**. Reject unknown keys, invalid values, and conflicting legacy/canonical duplicates, including invalid file values that environment overrides would otherwise mask.

<details>
<summary>Complete tocks configuration (all 48 defaults)</summary>

Merge this block into the Agency JSON config. Values are seconds; omitted settings use these defaults. Group aliases are omitted because every individual setting is shown.

```json
{
  "tocks": {
    "witnessMsg": 0.03125,
    "witnessEscrow": 0.03125,
    "witnessCue": 0.03125,
    "indirectorMsg": 0.03125,
    "indirectorCue": 0.03125,
    "indirectorEscrow": 0.03125,
    "mailboxPoll": 0.03125,
    "mailboxMsg": 0.03125,
    "mailboxEscrow": 0.03125,
    "pollerEvent": 0.03125,
    "pollerConnect": 0.03125,
    "receiptIntercept": 0.03125,
    "reactorMsg": 0.03125,
    "reactorCue": 0.03125,
    "reactorEscrow": 0.03125,
    "reactantMsg": 0.03125,
    "reactantCue": 0.03125,
    "reactantEscrow": 0.03125,
    "respondant": 0.03125,
    "receiptor": 0.03125,
    "oobi": 0.03125,
    "witnessReceiptor": 0.03125,
    "witnessReceiptorIdle": 0.03125,
    "witnessInquisitor": 0.03125,
    "witnessPublisher": 0.03125,
    "registrarEscrow": 0.03125,
    "credentialerEscrow": 0.03125,
    "counselorEscrow": 0.03125,
    "anchorerEscrow": 0.03125,
    "signify": {
      "initer": 0.03125,
      "querier": 0.03125,
      "escrower": 0.03125,
      "parser": 0.03125,
      "witnesser": 0.03125,
      "delegator": 0.03125,
      "exchangeSender": 0.03125,
      "granter": 0.03125,
      "admitter": 0.03125,
      "groupRequester": 0.03125,
      "seeker": 0.03125,
      "exchangecue": 0.03125,
      "submitter": 0.03125,
      "bootServer": 0.03125,
      "adminServer": 0.03125,
      "httpServer": 0.03125,
      "shutdown": 0.03125,
      "signalExpiry": 0.03125,
      "releaser": 60.0
    }
  }
}
```

</details>

## Scheduling and lifecycle

- Propagate worker cadences into transient query workers, witness catch-up/receipt generators, and exchange/grant delivery wrappers. Containers without independent settings inherit their parent's cadence.
- Make parser waits, delegation escrow, and shutdown loops yield their configured cadences. Keep Anchorer's escrow cadence separate from its container cadence.
- Schedule an Agent-owned `SignalExpirer` against the notifier's signal queue. Expiry scans a snapshot without yielding during iteration, avoiding deque mutation across yields while preserving the existing expiration timeout.
- Resolve Agency settings at startup and Agent settings at creation/reopen. New Agents persist explicit canonical file values, excluding environment overrides and defaults. Existing Agents reopen their own files without rewriting legacy configurations; there is no live reload or automatic synchronization from a changed Agency template.
- Align config-file lookup in `Agency.get()` with `Agency.create()` when Agent databases use a nonempty base.
- Extend both existing sample configs with all 48 settings and document defaults, aliases, precedence, lifecycle, and deprecation. One-off CLI commands and KERIpy's internal Poster, Clienter, and transport scheduling remain outside this change.

## Validation

Regression coverage includes configuration precedence and validation, legacy normalization and real-file reopening, database-base config lookup, independent Agent worker bindings, runtime cadence forwarding, signal expiry, and Agency service bindings. Sample-config tests check completeness and default values.

At commit `73fb151`, Linux and macOS test jobs (including lint and formatting), coverage, Docker build/start, and both Codecov checks pass.


src/keria/app/agenting.py
```diff
@@ -467,19 +465,21 @@ def get(self, caid):
         if aaid is None:
             return None
 
-        ks = keeping.Keeper(name=caid, base=self.base, temp=self.temp, reopen=True)
-
+        # Agent configs are written with base="" by _writeAgentConfig, even
+        # when the Agent's databases use a nonempty base.
         cf = configing.Configer(
-            name=caid, base=self.base, temp=self.temp, reopen=True, clear=False
```
kentbull, line 473: This aligns the read with the write side of thing. In `Agency._writeAgentConfig()` we do `base=""` as shown below, so we should do that on the read side to. If we want to change this that would be a later PR.

```py
class Agency(...):
    ...
    def _writeAgentConfig(self, caid):
        config = self._loadConfigForAgent(caid)
        cf = configing.Configer(
            name=f"{caid}",
            base="",
            human=False,
            temp=self.temp,
            reopen=True,
            clear=False,
        )
        cf.put(config)
        return cf
    ...
```

src/keria/app/agenting.py
```diff
@@ -1096,8 +1108,12 @@ def recur(self, tyme=None, tock=0.0, **opts):
             logger.info(
                 "Agent %s received:\n%s\n...\n", self.kvy, self.parser.ims[:1024]
             )
-        done = yield from self.parser.parsator()  # process messages continuously
-        return done  # should never get here except forced close
+        parsator = self.parser.parsator()
```
kentbull, line 1111: The `yield from self.parser.parsator()` did not respect the tock value passed in and effectively ran at `0.0`. Adding in the manual for loop with a yield ensures we respect the tock value.

kentbull:
@daidoji please review

src/keria/app/delegating.py
```diff
@@ -56,10 +57,17 @@ def __init__(self, hby, proxy=None, **kwa):
         self.postman = forwarding.Poster(hby=hby)
         self.witq = agenting.WitnessInquisitor(hby=hby)
         self.witDoer = agenting.Receiptor(hby=self.hby)
+        self.escrowDoer = doing.doify(self.escrowDo, tock=hby.tocks["anchorerEscrow"])
         self.proxy = proxy
 
+        kwa.setdefault("tock", scheduling.DEFAULT_TOCK)
```
iFergal, line 63: Why is this needed, if we are already reading directly from `hby.tocks` in `self.escrowDoer = doing.doify(self.escrowDo, tock=hby.tocks["anchorerEscrow"])`?

src/keria/app/agenting.py
```diff
@@ -323,6 +312,9 @@ def _loadConfigForAgent(self, caid):
         timestamp = nowIso8601()
         config = dict(self.cf.get() if self.cf is not None else {"dt": timestamp})
 
+        # Persist the configured snapshot, not environment overrides or defaults.
+        config["tocks"] = deepcopy(self._tockConfig.configured)
```
iFergal, line 316: Not sure I get this based on the PR description. `config["tocks"]` is written when the agent is created for a new Signify client, and this overwrites it every time with the agency `_tockConfig`?

src/keria/app/agenting.py
```diff
@@ -1590,15 +1616,33 @@ def recur(self, tyme, tock=0.0, **opts):
...
+    def __init__(self, signaler, tock=scheduling.DEFAULT_TOCK):
+        self.signaler = signaler
+        super().__init__(tock=tock)
+
+    def recur(self, tyme=None):
+        now = helping.nowUTC()
+        for sig in tuple(self.signaler.signals):
+            if now - helping.fromIso8601(sig.dt) > self.signaler.SignalTimeout:
+                self.signaler.signals.remove(sig)
+        return False
```
iFergal, line 1631: Could we also make a doer out of the `expireDoer` in the signaler?

src/keria/app/agenting.py
kentbull: `config["tocks"]` is initially set to the tocks loaded from the Agency's config file, which is what the `.configured` prop of `self._tockConfig` is. `self._tockConfig` is set to `scheduling.loadTocks(self.cf)`, which just loads the config file and resolves tock overrides. The tock overrides are in the `.signify` and `.keri` props of `_tockConfig`. The reason for the `.configured` prop of `_tockConfig` is to store the exact JSON config on disk, which becomes the Agent config.

The reason for this is that the Agency bootstraps the Agent config from the JSON config file, mirroring how we did things in the past, rather than having the tock overrides from environment variables be saved into the Agent config file.

We can change this. If we want the saved Agent config to be affected by the environment variable overrides then that would be easy to accommodate.

src/keria/app/agenting.py
kentbull: It would have been nice to do that, yet there's a risk that a future consumer of the `signals` queue sent into `Signaler` would either push or consume a signal, causing a possible `RuntimeError` like the following due to mutating the signals queue during iteration:

```py
RuntimeError: deque mutated during iteration
```

This is because `Signaler.expireDo` has `yield tock` in the body of the for loop iterating over the signals, which doesn't really make sense. It makes more sense to process all of the signals on a given recurrence of expireDo and then yield the tock.

```py
    def expireDo(self, tymth=None, tock=0.0, **kwa):
        ...
        while True:  # loop checking for expired messages
            now = helping.nowUTC()
            toRemove = []
            for sig in self.signals:
                if now - helping.fromIso8601(sig.dt) > self.SignalTimeout:  # Expire messages that are too old
                    toRemove.append(sig)
                yield tock  # <-- this is the problem, suspends this coroutine during queue iteration, exposing mutation errors

            for sig in toRemove:
                self.signals.remove(sig)

            yield tock
```

src/keria/app/delegating.py
kentbull: Setting a default of one tick (`DEFAULT_TOCK = 0.03125`) on every Doer/DoDoer/doified function in the codebase is why this line showed up.

The tock being set here is the `Anchorer` itself, not just the `escrowDoer` doified subtask. The `escrowDoer` has its own tock and the outer `Anchorer` DoDoer has its own, distinct tock that controls how often the DoDoer performs a recurrence processing once all the doers in its `.doers` prop.

The `kwa` is passed into the `super(...).__init__(...)` call as `**kwa`, which is where the tock gets passed to the DoDoer constructor to set its tock cadence.

src/keria/app/agenting.py
daidoji: Why do we yield the tock though as we iterate through the parsator?  Do we not care about the results yielded from that generator?

tests/scripts/keri/cf/main/keria.json
daidoji, line 1: Maybe we should name this file keria-legacy-config.json or something.  Not sure what else that will mess up but would be clear to future generations that this config style might disappear in the future.

daidoji (approved):

src/keria/app/agenting.py
kentbull: No parsed results are yielded outward to the scheduler from the `parsator`. The intention for a `.recur` method of a Doer is to yield a tock to the scheduler so it know when next to schedule the subtask.

Look through KERIpy's `Parser.parsator`. It delegates through yield from to `Parser.msgParsator, which in turn either yields nothing with bare yield statements. Even though `msgParsator` further delegates to `self._extractor` which does yield results to `msgParsator`, none of those yielded results are passed outward to the scheduler.

The reason this PR changed `ParserDoer.recur` to `yield self.tock` was so that this `Doer` behaved as the `Doist` scheduler expects where every subtask/coroutine (DoDoer, Doer, doified function) yields a tock defining the next time this particular subtask is asking the scheduler to schedule it. Previously, with bare yields this Doer was asking for whatever the parent scheduler's tock is, which is what HIO interprets a bare yield as, meaning it did not respect any configured tock of the subtask. 

HIO takes a bare yield and converts it to `None` which then is set by the scheduling context for a given Doer, which, if it is the Doist, usually defaults to 0.0.

We don't want that default here and instead want this Doer subtask to be able to configure its own tock, which is why we step once through the parsing loop and then yield a tock.

tests/scripts/keri/cf/main/keria.json
kentbull: I get where you're coming from, yet I see it differently. As this file contains only the new tock config I favor leaving it named as is. One way we could address your concern of showing just the legacy tocks would be to add a `keria-legacy-tocks.json` config file with only legacy tock config.

Or, we could use an `.hjson` file with comments since KERIpy's Configer class supports HJSON as well.

In a future PR :)

src/keria/app/delegating.py
iFergal: I understand that we are splitting the tocks for `Anchorer` and `escrowDoer`.

I'm just wondering why we are setting the `Anchorer` tock here when we already have `self.swain = delegating.Anchorer(hby=hby, proxy=agentHab, tock=parentTock)` in `agenting.py`

src/keria/app/agenting.py
iFergal: Should that be made as an upstream fix then?

src/keria/app/agenting.py
iFergal: Sorry I think my confusion here came from the naming of `_loadConfigForAgent` - this isn't loading an agent's config from the DB, this is called when creating the agent for the first time and loads the config into that agent's db. So this is fine then.

src/keria/app/delegating.py
kentbull: Huh, yes we could remove it. Setting it to the parent tock makes more sense. Will do.

src/keria/app/agenting.py
kentbull: Yes, eventually, though it takes long enough to fix upstream and get a release out that I don't want to delay this current release.

We could take it out of this PR, though having the fix here is nice.

kentbull: Agreed—the permanent fix belongs upstream. I opened https://github.com/WebOfTrust/keripy/issues/1690 with a reproduction and proposed fix.

KERIA currently pins KERIpy 1.2.14, whose `Signaler.expireDo` yields while iterating the live signal deque. Pushing a signal between those yields reproduces `RuntimeError: deque mutated during iteration`; wrapping that generator in a doer or changing its tock does not address the iterator invalidation.

The local `SignalExpirer` is a temporary workaround: it scans a snapshot and completes removals without yielding midway, preserving the existing expiration timeout and honoring KERIA's configured cadence. Once a KERIpy release includes the fix and KERIA adopts it, we should remove this class and schedule Signaler's expiry worker with the configured tock.

iFergal (approved):

